PR #18, opened 2020-05-15
add setvotebits api

dajohi:


router.go
```diff
@@ -15,6 +15,7 @@ func newRouter() *gin.Engine {
 		router.POST("/feeaddress", feeAddress)
 		router.GET("/pubkey", pubKey)
 		router.POST("/payfee", payFee)
+		router.POST("/setvotebits", setVoteBits)
```
jholdstock, line 18: Needs to be updated to `api.POST(...)`

methods.go
```diff
@@ -348,6 +348,70 @@ func PayFee2(ctx context.Context, ticketHash *chainhash.Hash, votingWIF *dcrutil
...
+
+	// TODO: DB - error if given timestamp is older than any previous requests
+
+	// TODO: DB - store setvotebits receipt in log
+
+	sendJSONResponse(setVoteBitsResponse{
+		Timestamp: time.Now().Unix(),
+		Request:   setVoteBitsRequest,
+		VoteBits:  voteBits,
+	}, http.StatusOK, c)
```
jholdstock, line 412: Need to remove `http.StatusOK`

PR #20, opened 2020-05-16
Add shutdown context.

jholdstock:
Code mostly taken from dcrwallet


jholdstock:
I am missing something here, the shutdown handlers run if I `CTRL+C`, but not if the process receives a `SIGINT`/`SIGTERM` - eg. from `kill` or `pkill`

jholdstock:
Fixed, ready for review.

PR #28, opened 2020-05-18
Error logging for API methods

jholdstock:
- Use gin for unmarshalling json
- Logging for every negative path in every API method. Bad requests are `WARN`, internal server errors are `ERROR`
- Removed explicit checks for `MaxHashStringSize` because this is already checked in `NewHashFromStr`


dajohi:
NewHashFromStr only checks if larger than MaxHashStringSize.

PR #86, opened 2020-05-28
Voting wallet consistency checking

### LINKED ISSUE #46: Consistency checking
Inconsistencies between voting wallets may arise from:

- A new wallet being added to the deployment
- RPC errors

We need to ensure:
- All voting wallets are aware of all tickets
- All voting wallets have the private keys to vote on all tickets
- Vote Choices are correct for all tickets

We probably want to run consistency checks when dcrvsp is started up, whenever we are aware an RPC error has occured (maybe start a timer?). We could also add the ability to manually trigger a consistency check - HTTP endpoint perhaps?

### LINKED ISSUE #103: [voting wallet] Warn if wallet stop
Right now the vpsd won't allow to start without a dcrwallet unlocked and with `--enablevoting`. But if dcrwallet crashes or stop working for some reason, the vspd don't even warn about it.

I think it would be good having some notification for that

jholdstock:
This PR makes use of the `ticketinfo` RPC [added to dcrwallet](https://github.com/decred/dcrwallet/pull/1783) to:

- Perform voting wallet consistency checks
- Set ticket outcome in the database

---

#### Consistency checking

Consistency checks are run immediately on vspd startup and then once every 30 minutes (Closes #46). Consistency checking is a two step process:

1. Query the database for all tickets which should currently be added to voting wallets, and use `ticketinfo` on each voting wallet to ensure all of the wallets are aware of all of the tickets
1. Ensure every ticket on every wallet has the correct voting choices set

---

#### Set ticket outcome

Using `ticketinfo` allows vspd to determine when a ticket is no longer votable, either because it has already voted or because it has been revoked. This information is now stored in the vspd database. The `ticketinfo` RPC does not distinguish between missed and expired tickets, and so vspd simply buckets both of these into `revoked`.

This occurs each time a block is connected by the local dcrd instance, which provides a regular check on the status of dcrwallet RPC connections (Closes #103).

Counts of `Voting`/`Voted`/`Revoked` tickets are now reported on both the web front-end and through the `/vspinfo` API call. This replaces the previous counts of `Total tickets` and `Fee confirmed tickets`. Tickets with unconfirmed fees are no longer reported.


jholdstock:
I believe this is ready to go, but its rebased on #101. Will mark for review when 101 is in.

jholdstock:
Ran into an issue.

- `gettickets` RPC only returns tickets which are live.
- vspd does not know if a ticket has voted or not.

When a ticket is missing from `gettickets` response, vspd cannot know whether this is because the wallet does not know about the ticket, or because the ticket has voted.

This means that the current code is continuously trying to add tickets which have already voted to the voting wallets.

jholdstock:
Ready to go, just waiting for the wallet PR to be merged

PR #93, opened 2020-06-01
Reject unvotable tickets.

### LINKED ISSUE #56: Ensure tickets are either immature or live
There is currently a primitive check in `/feeaddress` to ensure a ticket is not older than `TicketMaturity+TicketExpiry`, but we should be able to validate tickets a bit more carefully.

We should not issue fee addresses, or broadcast fee transactions, for tickets which are expired, missed or revoked.

jholdstock:
/payfee and /getaddress will now only accept tickets which are immature or live.

Closes #56

### DIFF
--- a/rpc/dcrd.go
+++ b/rpc/dcrd.go
@@ -162,8 +162,9 @@ func (c *DcrdRPC) CanTicketVote(ticketHash string, netParams *chaincfg.Params) (
 		return false, err
 	}
 
-	// Tickets which older than (TicketMaturity+TicketExpiry) are too old to vote.
-	if rawTx.Confirmations > int64(uint32(netParams.TicketMaturity)+netParams.TicketExpiry) {
+	// Tickets which have more than (TicketMaturity+TicketExpiry+1)
+	// confirmations are too old to vote.
+	if rawTx.Confirmations > int64(uint32(netParams.TicketMaturity)+netParams.TicketExpiry)+1 {
 		return false, nil
 	}
 
--- a/rpc/dcrd.go
+++ b/rpc/dcrd.go
@@ -10,6 +10,7 @@ import (
 	"github.com/decred/dcrd/chaincfg/v3"
 	dcrdtypes "github.com/decred/dcrd/rpc/jsonrpc/types/v2"
 	"github.com/decred/dcrd/wire"
+	"github.com/jrick/bitset"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


rpc/dcrd.go
```diff
@@ -135,3 +136,47 @@ func (c *DcrdRPC) GetBestBlockHeader() (*dcrdtypes.GetBlockHeaderVerboseResult,
...
+	if err != nil {
+		return false, err
+	}
+
+	return bitset.Bytes(existsBytes).Get(0), nil
+}
+
+// CanTicketVote checks determines whether a ticket is able to vote at some
+// point in the future by checking that it is currently either immature or live.
+func (c *DcrdRPC) CanTicketVote(ticketHash string, netParams *chaincfg.Params) (bool, error) {
```
jholdstock, line 158: @davecgh could you give this function a once-over please?

rpc/dcrd.go
```diff
@@ -135,3 +136,47 @@ func (c *DcrdRPC) GetBestBlockHeader() (*dcrdtypes.GetBlockHeaderVerboseResult,
...
+	// Get ticket details.
+	rawTx, err := c.GetRawTransaction(ticketHash)
+	if err != nil {
+		return false, err
+	}
+
+	// Tickets which older than (TicketMaturity+TicketExpiry) are too old to vote.
```
davecgh, line 165: missing `are`.

rpc/dcrd.go
```diff
@@ -135,3 +136,47 @@ func (c *DcrdRPC) GetBestBlockHeader() (*dcrdtypes.GetBlockHeaderVerboseResult,
...
+	if rawTx.Confirmations > int64(uint32(netParams.TicketMaturity)+netParams.TicketExpiry) {
```
davecgh, line 166: This looks off by one to me.  Tickets didn't become live until the block _after_ they reach ticket maturity.

This can be seen by noticing that the first tickets were mined in [block 257](https://explorer.dcrdata.org/block/000000000001854541511c2de81797bc2e7e03990b7f84920c68375159c500ad) and don't become live until 257+256+1 = [514](https://explorer.dcrdata.org/block/000000000000402d8cffaadbde7d866517051130e44549df58d2b8ec0567eab5) by looking at the fact the pool size becomes nonzero at that point.

Then, those ticket would expire 40960 blocks after the point they became live.  Ergo, a ticket mined in block 257 that didn't get selected would expire in block 514+ 40960 = [41474](https://explorer.dcrdata.org/block/000000000000009845d6f8469b484c35e34b1fc92b9fa39c54588703e0a72b41), and as luck would have it for this example, [one of those tickets](https://explorer.dcrdata.org/tx/9043be520f3f79d9e134ad43dd3bb5a3d41d5ba46c4624846f0c6648123c5991) actually did end up [expiring and was revoked in the aforementioned block](https://explorer.dcrdata.org/tx/242d7bb7dc0c67395e27f8dfb8f28cb46dbd55b36c34d22451bbed76e28bfc27).

So, running the numbers with what it would look like at block 41473 would give:

Confirmations = 41473 - 257 (block it was mined in) + 1 (including the block it was mined in) = 41217.
TicketMaturity + TicketExpiry = 40960 + 256 = 41216

Then `if 41217 > 41216` is true, which is incorrect, because it wouldn't expire until the next block.

jholdstock:
Thanks for explaining that Dave, I've added a `+1` to accommodate.

Issue #88, opened 2020-05-28
Deployment guide


jholdstock:
`README.md` only briefly describes how to effectively deploy vspd. A thorough deployment guide should be written, in line with what is provided for dcrstakepool.

- https://github.com/decred/dcrstakepool/blob/master/README.md
- https://docs.decred.org/advanced/operating-a-vsp/

We may even want two docs here, one for new vspd deployments, and one for deploying vspd alongside an existing dcrstakepool deployment.

Issue #66, opened 2020-05-26
Idea: store fee xpub in database


jholdstock:
Currently dcrvsp is loading a fee xpub from config every time it starts. I have two issues with this:

- If the operator changes the value set in config, the behaviour is undefined.
- It adds an extra step to recovering a backup. The operator has to put their backed up database in place, and also ensure the correct xpub is set in config.

Alternative: Accept an xpub once, first time a VSP is started, probably via CLI flag. Store this in the database. This fixes both issues above. Comparable to having to init `dcrwallet` the first time it is run, using `--create`

PR #95, opened 2020-06-02
Add deployment guide.

jholdstock:
Implements most of #88, but monitoring and backup/restore are still not done. 

Would like to decide if we are implementing #66 before completing the backup/restore docs.

For monitoring, we may want to add some http endpoint for pinging the front end. @dajohi do you still have monit configurations for dcrstakepool? Could be adapted for vspd?


jholdstock:
@vctt94 could you confirm if this deployment guide covers everything you encountered when setting up vspd?

PR #114, opened 2020-06-05
remove fee transactions if dcrd rejects it.

dajohi:


webapi/payfee.go
```diff
@@ -216,7 +216,13 @@ findAddress:
 		feeTxHash, err := dcrdClient.SendRawTransaction(payFeeRequest.FeeTx)
 		if err != nil {
 			log.Errorf("SendRawTransaction failed: %v", err)
+			log.Infof("Removing fee transaction for ticket %v", ticket.Hash)
 			sendErrorResponse("dcrwallet RPC error", http.StatusInternalServerError, c)
+			ticket.FeeTxHex = ""
```
jholdstock, line 221: I think this needs to be reshuffled a little bit. We shouldn't set the feetx in the database and then un-set it. If the ticket is already confirmed, and broadcasting the fee tx fails, then we actually don't need to update the database at all - we dont need to store the wif, the hex or the vote choices. Just return an error and let the client try again with a new tx.

jholdstock:
We should document this behaviour. Suggest adding to `### Ticket Status` section of `docs/api.md`:

```
- `feetxreceived` is true when the VSP has received a valid fee transaction. If vspd encounters an error when broadcasting the fee tx, this will flip back to false.
```

PR #150, opened 2020-06-30
database: protect access with a mutex

dajohi:


database/ticket.go
```diff
@@ -184,6 +199,9 @@ func (vdb *VspDatabase) CountTickets() (int, int, error) {
 
 // GetUnconfirmedTickets returns tickets which are not yet confirmed.
 func (vdb *VspDatabase) GetUnconfirmedTickets() ([]Ticket, error) {
+	defer vdb.dbMtx.RUnlock()
```
jholdstock, line 202: I guess this can be removed from here and added to `filterTickets()`?  Same comment for other funcs using`filterTickets()`.

No need for the duplication

dajohi: I prefer to keep it as-is.  If any exported funcs ever need to do more with the results while keeping the lock, it will race or deadlock.

I added a comment to filterTickets that it needs to be called with the lock held.

PR #154, opened 2020-07-13
Fix ticket status request 

vctt94:
Right now when sending a ticket status request, we are receiveing a EOF error.

This PR fixes it.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -145,7 +145,7 @@ If `feetxstatus` is `error`, the client needs to provide a new fee transaction
 using `/api/payfee`. The VSP will only add a ticket to the voting wallets once
 its `feetxstatus` is `confirmed`.
 
-- `GET /api/ticketstatus`
+- `POST /api/ticketstatus`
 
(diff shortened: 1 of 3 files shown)


jholdstock:
I want to look further at this issue because the fix in this PR does not make it clear exactly what the problem was. Adding timestamp to the `ticketHashRequest` really should not be necessary afaict.

I'm not able to recreate the issue. I am sending a signed HTTP request like the following:

```
GET /api/ticketstatus

VSP-Client-Signature: Hyw2mgFS15Q4lFTwe0zWze+/dllZcNbFGhUt6zdgibyFD0HSae3XnkEPSBM126JmQXKEG8P6eWvyZ5vwfPeGxlU=

{"timestamp":1594719081,"tickethash":"e96252fbd73d4bbfadd2a4551d3ce28bbb49c97aff7e3cae79aea5810bd397c2"}
```

And response is coming back as expected.

Could you please share your problematic request, with the same details I posted above?

PR #156, opened 2020-07-14
API versioning

### LINKED ISSUE #118: [url] Add API version into URL request
Older versions of vspd have the API version into the request, for example: `/api/v1/getpurchaseinfo`, but right now this new version does not have it, example: `/api/vspinfo`.

We should call it `/api/v3/vspinfo` to keep consistency.

jholdstock:
Adds `/v1`/ to API routes, eg `/api/v1/ticketstatus`

`/api/vspinfo` remains on the same path as before, but a new field `apiversions` has been added, so clients are able to check supported API versions before using the API.

Closes #118

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -21,6 +21,11 @@
 - Implementation of request and response types can be found in
   [webapi/types.go](../webapi/types.go).
 
+- The initial version of the vspd API is version 3. This is because the first
+  version of the vspd API actually represents the third iteration of VSP APIs.
+  The first and second iterations of VSP API were implemented by
+  [dcrstakepool](https://github.com/decred/dcrstakepool).
+
 ## Expected usage
 
 ### Get VSP info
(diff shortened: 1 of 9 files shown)


jholdstock:
Rebased, updated to use `/v3` instead of `/v1`, and added some notes to `deployment.md` and `api.md`

docs/api.md
```diff
@@ -21,24 +21,30 @@
 - Implementation of request and response types can be found in
   [webapi/types.go](../webapi/types.go).
 
+- The initial version of the vspd API is numbered 3. This is because the first
+  version of the vspd API actually represents the third iteration of VSP APIs.
+  The first and second iterations of VSP API were implemented by
+  [dcrstakepool](https://github.org/decred/dcrstakepool).
```
dajohi, line 27: github.com

jholdstock: ffs, second time I've made that mistake recently. I've somehow gotten `github.org` lodged into muscle memory. I've fixed that, and rebased onto master.

PR #167, opened 2020-08-03
Add basic http auth for /admin/status. Update docs.

### LINKED ISSUE #102: Monitoring guide
Do we want to add an HTTP endpoint for pinging the front end?

@dajohi do you still have monit configurations for dcrstakepool? Could be adapted for vspd?

jholdstock:
Closes #102

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -77,6 +77,24 @@ dcrstakepool deployment.
 
 ## Monitoring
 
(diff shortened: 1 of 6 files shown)


jholdstock:
I've pushed an extra commit here which fixes an issue where admin pages would not load if there were no voting wallets connected.

PR #169, opened 2020-08-07
Write a backup after closing the db, not before

jholdstock:
If a backup is written before the database is closed, there may be database transactions which are left out of the backup.

This PR moves the creation of the backup so it occurs **after** the db is closed. This should ensure the backup always reflects the latest state of the db.

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"encoding/binary"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


database/database.go
```diff
@@ -179,18 +194,62 @@ func Open(ctx context.Context, shutdownWg *sync.WaitGroup, dbFile string, backup
...
 	if err != nil {
-		log.Errorf("Error closing database: %v", err)
-	} else {
-		log.Debug("Database closed")
+		log.Errorf("Failed to write a database backup (os.Open): %v", err)
+		return
 	}
+	defer from.Close()
+
+	to, err := os.OpenFile(tempPath, os.O_RDWR|os.O_CREATE, 0666)
```
dajohi, line 232: 0600

jholdstock: Good catch - fixed

PR #172, opened 2020-08-17
Update to go 1.15

jholdstock:


jholdstock:
@dajohi looks like this might need some admin magic to merge to master without passing the 1.13 check

PR #176, opened 2020-08-18
Only load .html files in template dir

jholdstock:
To fix an issue seen by @isuldor

vspd does not need to parse non-html files in the template dir (eg. temporary/hidden files created by text editors)


isuldor:
I confirmed this fixes the issue I was observing, thanks!

dajohi (approved):

PR #177, opened 2020-08-18
Return descriptive error if ticket cannot be broadcast

### LINKED ISSUE #171: [error] vspd returns internal error when should be more informative
Right now, if we have a failure in send transactions, for example, vspd will return an internal error: 

```
2020-08-14 14:23:03.481 [WRN] VSPC: vsp responded with an error: {"code":1,"message":"internal error"}
2020-08-14 14:23:03.481 [WRN] VSPC: Failed to process queued ticket a1aeb0764e28cf76b74714ed8be0c6c306058880a1808a4639cbcb80708c83b3
```

While the true error was:

``` 
2020-08-14 17:23:03.429 [ERR] API: broadcastTicket: dcrd.SendRawTransaction for ticket failed (ticketHash=a1aeb0764e28cf76b74714ed8be0c6c306058880a1808a4639cbcb80708c83b3): rejected transaction a1aeb0764e28cf76b74714ed8be0c6c306058880a1808a4639cbcb80708c83b3: orphan transa
ction a1aeb0764e28cf76b74714ed8be0c6c306058880a1808a4639cbcb80708c83b3 references outputs of unknown or fully-spent transaction 50d28c70ef3a9d94b1cd979ba7b388c6a44fa692beb7fe251f4c13fbdffa069e
```

We should return the error in this case, so we can inform the user what to do, like a rescan, for example.

jholdstock:
Closes #171

### DIFF
--- a/webapi/errors.go
+++ b/webapi/errors.go
@@ -19,6 +19,7 @@ const (
 	errInvalidPrivKey
 	errFeeNotReceived
 	errInvalidTicket
+	errCannotBroadcastTicket
 )
 
 // httpStatus maps application error codes to HTTP status codes.
@@ -52,6 +53,8 @@ func (e apiError) httpStatus() int {
 		return http.StatusBadRequest
 	case errInvalidTicket:
 		return http.StatusBadRequest
+	case errCannotBroadcastTicket:
+		return http.StatusInternalServerError
 	default:
 		return http.StatusInternalServerError
 	}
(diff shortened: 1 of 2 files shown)


dajohi:
Can you do one for the fee transaction as well?

dajohi (changes requested):
Add fee transaction broadcast error as well

PR #184, opened 2020-09-07
Tidy up shutdown logic

### LINKED ISSUE #173: RPC client WaitGroup panic
```
2020-08-17 15:56:22.564 [DBG] RPC: RPC client wss://127.0.0.1:19110/ws errored (read tcp 127.0.0.1:45924->127.0.0.1:19110: use of closed network connection); reconnecting...
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 1 [running]:
sync.(*WaitGroup).Wait(0xc000277850)
        sync/waitgroup.go:132 +0xad
main.run(0xdccb20, 0xc000285500, 0x0, 0x0)
        github.com/decred/vspd@/vspd.go:106 +0xa68
main.main()
        github.com/decred/vspd@/vspd.go:25 +0x81
```

jholdstock:
Removing the use of a global context and global WaitGroup in `background.go`. Should close #173 

Background processes triggered by timers were previously adding 1 to the shutdown WaitGroup each time they were triggered, and subtracting 1 each time they completed. This resulted in #173, where the WaitGroup was incremented incorrectly. Now - the WaitGroup is incremented once when the timer is started, and once when it has stopped. 

This change - as well as using a background context for these background tasks - also helps shutdown to be more robust. Now the background processes will terminate before shutdown proceeds, whereas previously they could be interrupted.

Another added benefit is that the global copies of the WaitGroup and the shutdown context in `background.go` are removed by this PR.

### DIFF
--- a/background/background.go
+++ b/background/background.go
@@ -279,7 +279,7 @@ func (n *NotificationHandler) Close() error {
 func connectNotifier(shutdownCtx context.Context, dcrdWithNotifs rpc.DcrdConnect) error {
 	notifierClosed = make(chan struct{})
 
-	dcrdClient, err := dcrdWithNotifs.Client(context.Background(), netParams)
+	dcrdClient, err := dcrdWithNotifs.Client(shutdownCtx, netParams)
 	if err != nil {
 		return err
 	}
@@ -295,6 +295,9 @@ func connectNotifier(shutdownCtx context.Context, dcrdWithNotifs rpc.DcrdConnect
 	// notifier is closed.
 	select {
 	case <-shutdownCtx.Done():
+		// A shutdown signal has been received - close the client with the
+		// notification handler to prevent further notifications from being
+		// received.
 		dcrdWithNotifs.Close()
 		return nil
 	case <-notifierClosed:
--- a/background/background.go
+++ b/background/background.go
@@ -19,13 +19,11 @@ import (
 )
 
 var (
-	ctx            context.Context
 	db             *database.VspDatabase
 	dcrdRPC        rpc.DcrdConnect
 	walletRPC      rpc.WalletConnect
 	netParams      *chaincfg.Params
 	notifierClosed chan struct{}
-	shutdownWg     *sync.WaitGroup
 )
 
 type NotificationHandler struct{}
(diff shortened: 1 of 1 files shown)


background/background.go
```diff
@@ -324,37 +320,45 @@ func Start(c context.Context, wg *sync.WaitGroup, vdb *database.VspDatabase, drp
...
+	consistencyLoop:
 		for {
 			select {
-			case <-ctx.Done():
+			case <-shutdownCtx.Done():
 				ticker.Stop()
-				return
+				break consistencyLoop
 			case <-ticker.C:
 				checkWalletConsistency()
```
dajohi, line 333: could pass shutdownCtx to checkWalletConsistency()

background/background.go
```diff
@@ -279,10 +276,10 @@ func (n *NotificationHandler) Close() error {
 	return nil
 }
 
-func connectNotifier(dcrdWithNotifs rpc.DcrdConnect) error {
+func connectNotifier(shutdownCtx context.Context, dcrdWithNotifs rpc.DcrdConnect) error {
 	notifierClosed = make(chan struct{})
 
-	dcrdClient, err := dcrdWithNotifs.Client(ctx, netParams)
+	dcrdClient, err := dcrdWithNotifs.Client(context.Background(), netParams)
```
dajohi, line 282: pass shutdownCtx

background/background.go
```diff
@@ -66,8 +64,7 @@ func blockConnected() {
```
dajohi, line 64: can blockConnected() take shutdownCtx?

background/background.go
jholdstock: I don't think that would work well because the shutdown context is cancelled as soon as CTRL+c (or another shutdown signal) is received, and wsrpc clients stop working when their context is cancelled.

If a shutdown signal is received while `checkWalletConsistency()` is running, RPC clients will throw errors and the consistency checking process will error out before completing.

background/background.go
jholdstock: Using the shutdown context is acceptable here because this client is only used once to register for notifications. It should never be used after a shutdown signal has been received.

I will update the PR.

background/background.go
jholdstock: See comment on `checkWalletConsistency()` - we want the RPC clients to be available after a shutdown signal is received so an in-progress `blockConnected()` func can complete its work before vspd shuts down.

dajohi (approved):

PR #189, opened 2020-10-01
rpc: optimize TicketInfo

dajohi:
This changes TicketInfo to only scan from the block height
required to reconcile.

Also, update GetBestBlockHeight to just use getblockcount

### DIFF
--- a/background/background.go
+++ b/background/background.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"strings"
 	"sync"
 	"time"
 
@@ -209,9 +210,19 @@ func blockConnected() {
 				for agenda, choice := range ticket.VoteChoices {
 					err = walletClient.SetVoteChoice(agenda, choice, ticket.Hash)
 					if err != nil {
-						log.Errorf("%s: dcrwallet.SetVoteChoice error (wallet=%s, ticketHash=%s): %v",
-							funcName, walletClient.String(), ticket.Hash, err)
-						continue
+						if strings.Contains(err.Error(), "no agenda with ID") {
+							log.Warnf("%s: Removing invalid agenda from ticket vote choices (ticketHash=%s, agenda=%s)",
+								funcName, ticket.Hash, agenda)
+							delete(ticket.VoteChoices, agenda)
+							err = db.UpdateTicket(ticket)
+							if err != nil {
+								log.Errorf("%s: db.UpdateTicket error, failed to remove invalid agenda (ticketHash=%s): %v",
+									funcName, ticket.Hash, err)
+							}
+						} else {
+							log.Errorf("%s: dcrwallet.SetVoteChoice error (wallet=%s, ticketHash=%s): %v",
+								funcName, walletClient.String(), ticket.Hash, err)
+						}
 					}
 				}
 				log.Infof("%s: Ticket added to voting wallet (wallet=%s, ticketHash=%s)",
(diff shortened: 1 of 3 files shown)


davecgh (approved):

rpc/dcrwallet.go
```diff
@@ -181,22 +181,22 @@ func (c *WalletRPC) SetVoteChoice(agenda, choice, ticketHash string) error {
...
 }
 
-// GetBestBlockHeight uses getbestblock RPC to query the height of the best
+// GetBestBlockHeight uses getblockcount RPC to query the height of the best
 // block known by the dcrwallet instance.
 func (c *WalletRPC) GetBestBlockHeight() (int64, error) {
-	var block dcrdtypes.GetBestBlockResult
-	err := c.Call(c.ctx, "getbestblock", &block)
+	var height int64
+	err := c.Call(c.ctx, "getblockcount", &height)
```
jholdstock, line 188: Is the switch from getbestblock to getblockcount just for the sake of efficiency? Or is there a functional difference?

database/ticket.go
```diff
@@ -43,6 +43,7 @@ const (
 // deliberately kept short because they are duplicated many times in the db.
 type Ticket struct {
 	Hash              string `json:"hsh"`
+	BlockHeight       int64  `json:"bhgt"`
```
jholdstock, line 46: Can we go with `PurchaseHeight` to make the name slightly more descriptive please

background/background.go
```diff
@@ -209,6 +211,14 @@ func blockConnected() {
 					if err != nil {
```
jholdstock, line 211: Because we do expect this scenario to happen occasionally, and we have appropriate handling for it, I think we should avoid logging errors and just log some warnings instead.

```
err = setVoteChoice
if error {
    if "no agenda with ID" {
        log.warn "Removing invalid agenda from ticket vote choices (ticketHash=blah, agenda=blah)"
        delete choice
    } else {  // any other error
       log.error
    }
}
```

background/background.go
```diff
@@ -209,6 +211,14 @@ func blockConnected() {
...
 							funcName, walletClient.String(), ticket.Hash, err)
+						if strings.Contains(err.Error(), "no agenda with ID") {
+							delete(ticket.VoteChoices, agenda)
+							err = db.UpdateTicket(ticket)
+							if err != nil {
+								log.Errorf("%s: db.UpdateTicket error, failed to remove invalid agenda (ticketHash=%s): %v",
+									funcName, ticket.Hash, err)
+							}
+						}
 						continue
```
jholdstock, line 222: This `continue` seems to be redundant, I think it can be removed.

background/background.go
```diff
@@ -232,7 +242,18 @@ func blockConnected() {
 			continue
 		}
 
-		ticketInfo, err := walletClient.TicketInfo()
+		// Find the oldest block height from confirmed tickets.
+		var startHeight int64
```
jholdstock, line 246: Extract this to a func rather than copy/pasting please.

```
func findOldestHeight(tickets []database.Ticket) int64 {
    ...
}
```

background/background.go
```diff
@@ -501,6 +533,15 @@ func checkWalletConsistency() {
 				if err != nil {
 					log.Errorf("%s: dcrwallet.SetVoteChoice error (wallet=%s, ticketHash=%s): %v",
 						funcName, walletClient.String(), dbTicket.Hash, err)
+					if strings.Contains(err.Error(), "no agenda with ID") {
```
jholdstock, line 536: Should follow same pattern described above for logging.

jholdstock (changes requested):

jholdstock (approved):

PR #193, opened 2020-10-11
fix errorlint warnings

dajohi:


config.go
```diff
@@ -183,11 +183,14 @@ func loadConfig() (*config, error) {
 	preParser := flags.NewParser(&preCfg, flags.HelpFlag)
 	_, err := preParser.Parse()
 	if err != nil {
-		if e, ok := err.(*flags.Error); ok && e.Type != flags.ErrHelp {
-			return nil, err
-		} else if ok && e.Type == flags.ErrHelp {
-			fmt.Fprintln(os.Stdout, err)
-			os.Exit(0)
+		var e *flags.Error
```
jholdstock, line 186: Please double check, but I think this can be simplified to the following?

```
	if err != nil {
		var e *flags.Error
		if errors.As(err, &e) {
			if e.Type == flags.ErrHelp {
				fmt.Fprintln(os.Stdout, err)
				os.Exit(0)
			}
		}
		return nil, err
	}
```

jholdstock: Maybe even this?

```
	if err != nil {
		var e *flags.Error
		if errors.As(err, &e) && e.Type == flags.ErrHelp {
			fmt.Fprintln(os.Stdout, err)
			os.Exit(0)
		}
		return nil, err
	}
```

jholdstock: This is still too verbose. The `else if` condition is not necessary. Equality between `e.Type` and `flags.ErrHelp` has already been checked.

See my suggestion above.

dajohi: agree

jholdstock (approved):

PR #205, opened 2020-11-18
api: add optional parent transaction to feeaddress

dajohi:


webapi/middleware.go
```diff
@@ -25,8 +25,9 @@ type ticketHashRequest struct {
 }
 
 type ticketRequest struct {
-	TicketHex  string `json:"tickethex" binding:"required"`
-	TicketHash string `json:"tickethash" binding:"required"`
+	TicketHex  string  `json:"tickethex" binding:"required"`
+	TicketHash string  `json:"tickethash" binding:"required"`
+	ParentHex  *string `json:"parenthex,omitempty"`
```
jholdstock, line 30: `omitempty` only effects marshalling, not unmarshalling, and thus it isn't needed on requests, only responses.

A field is made optional by omitting `binding:"required"`

webapi/types.go
```diff
@@ -18,9 +18,10 @@ type vspInfoResponse struct {
 }
 
 type feeAddressRequest struct {
-	Timestamp  int64  `json:"timestamp" binding:"required"`
-	TicketHash string `json:"tickethash" binding:"required"`
-	TicketHex  string `json:"tickethex" binding:"required"`
+	Timestamp  int64   `json:"timestamp" binding:"required"`
+	TicketHash string  `json:"tickethash" binding:"required"`
+	TicketHex  string  `json:"tickethex" binding:"required"`
+	ParentTx   *string `json:"parenthex,omitempty"`
```
jholdstock, line 24: `omitempty` not required

webapi/middleware.go
```diff
@@ -162,28 +163,68 @@ func broadcastTicket() gin.HandlerFunc {
...
-		// hex, so we can broadcast it here.
-		var e *wsrpc.Error
-		if errors.As(err, &e) && e.Code == rpc.ErrNoTxInfo {
+		case errors.As(err, &e) && e.Code == rpc.ErrNoTxInfo:
+			// ErrNoTxInfo means local dcrd is not aware of the ticket. We have the
+			// hex, so we can broadcast it here.
 			log.Debugf("%s: Broadcasting ticket (ticketHash=%s)", funcName, request.TicketHash)
 			err = dcrdClient.SendRawTransaction(request.TicketHex)
-			if err != nil {
+			switch {
```
jholdstock, line 216: Duplicate tx errors already handled in the RPC client https://github.com/decred/vspd/blob/0f657eb4fe6b143d757a5cd2936b7d04cef6849e/rpc/dcrd.go#L121-L158

webapi/middleware.go
```diff
@@ -162,28 +163,68 @@ func broadcastTicket() gin.HandlerFunc {
...
+				break
+			}
+			if !found {
+				log.Errorf("%s: Invalid ticket parent (ticketHash=%s)", funcName, request.TicketHash)
+				sendErrorWithMsg("invalid ticket parent", errBadRequest, c)
+				return
+			}
+			log.Debugf("%s: Broadcasting parent tx %s (ticketHash=%s)", funcName, parentHash, request.TicketHash)
+			err = dcrdClient.SendRawTransaction(request.TicketHex)
+			switch {
```
jholdstock, line 191: Duplicate tx errors already handled in the RPC client

webapi/middleware.go
jholdstock: Also lets stay consistent with the rest of the fields and use a string here rather than a pointer. Otherwise it isn't immediately obvious whether the check for existence should be `!= nil`, `!= ""`, or both.

webapi/middleware.go
```diff
@@ -162,28 +163,68 @@ func broadcastTicket() gin.HandlerFunc {
 		}
 
 		dcrdClient := c.MustGet("DcrdClient").(*rpc.DcrdRPC)
+		var e *wsrpc.Error
+
+		if request.ParentHex != nil {
+			parentTx, err := decodeTransaction(*request.ParentHex)
+			if err != nil {
+				log.Errorf("%s: Failed to decode parent hex (ticketHash=%s): %v", funcName, request.TicketHash, err)
```
jholdstock, line 171: Hmm I guess we don't actually need to fail and return an error here. There is a chance the provided ticket tx could be valid even if the parent tx is not.

We could still attempt to validate/broadcast the ticket tx, and if that fails, return an error at that point.

```
// If this errors, just log an error and continue. There is a chance the provided ticket tx could be valid even if the parent tx is not.
log.Warnf("%s: Failed to decode parent hex (ticketHash=%s): %v", funcName, request.TicketHash, err)

**Review by jholdstock — CHANGES_REQUESTED** [review]
Need to update docs https://github.com/decred/vspd/blame/master/docs/api.md#L74-L83

File: webapi/middleware.go (hunk 5 above)
Comment (dajohi) [reply webapi/middleware.go:170]: I think if it is supplied it should be valid to prevent using the vsp to send any tx

File: webapi/middleware.go [hunk 6]
```diff
@@ -163,6 +164,37 @@ func broadcastTicket() gin.HandlerFunc {
...
+				found = true
+				break
+			}
+			if !found {
+				log.Errorf("%s: Invalid ticket parent (ticketHash=%s)", funcName, request.TicketHash)
+				sendErrorWithMsg("invalid ticket parent", errBadRequest, c)
+				return
+			}
+			log.Debugf("%s: Broadcasting parent tx %s (ticketHash=%s)", funcName, parentHash, request.TicketHash)
+			err = dcrdClient.SendRawTransaction(request.TicketHex)
```
jholdstock, line 189: We already do this, look down :point_down:

jholdstock (changes requested):

webapi/middleware.go
```diff
@@ -163,6 +164,37 @@ func broadcastTicket() gin.HandlerFunc {
 
 		dcrdClient := c.MustGet("DcrdClient").(*rpc.DcrdRPC)
 
+		if request.ParentHex != "" {
+			parentTx, err := decodeTransaction(request.ParentHex)
```
jholdstock, line 168: Maybe `getTransaction` for the parent first? If it exists, we can skip the validation and SendRawTransaction.

jholdstock (approved):

PR #214, opened 2020-12-22
Only marshal DB record when needed

jholdstock:
FYI @dnldd

### DIFF
--- a/database/ticket.go
+++ b/database/ticket.go
@@ -82,8 +82,8 @@ var (
 )
 
 func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
-	defer vdb.ticketsMtx.Unlock()
 	vdb.ticketsMtx.Lock()
+	defer vdb.ticketsMtx.Unlock()
 
 	return vdb.db.Update(func(tx *bolt.Tx) error {
 		ticketBkt := tx.Bucket(vspBktK).Bucket(ticketBktK)
@@ -126,8 +126,8 @@ func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
 }
 
 func (vdb *VspDatabase) DeleteTicket(ticket Ticket) error {
-	defer vdb.ticketsMtx.Unlock()
 	vdb.ticketsMtx.Lock()
+	defer vdb.ticketsMtx.Unlock()
 
 	return vdb.db.Update(func(tx *bolt.Tx) error {
 		ticketBkt := tx.Bucket(vspBktK).Bucket(ticketBktK)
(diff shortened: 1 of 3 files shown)


database/ticket.go
```diff
@@ -82,6 +82,13 @@ var (
 )
 
 func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
+	// Marshal ticket first. This could prevent a DB transaction being started
+	// and rolled back unnecessarily.
+	ticketBytes, err := json.Marshal(ticket)
```
dnldd, line 87: Hmm, could be said that the ticket is also being marshalled too early looking at where its used. Also isn't it highly unlikely this the marshal fails? If that's the case then the possibility of a rollback would be minimal.

dnldd: A place I think we could look for some performance gains is iterating the ticket bucket to check if there is a ticket with the same fee address. My thinking is if the fee address and index are part of the key of the ticket then it can be extracted from it and used in the comparison and there would be no need to unmarshal the ticket value just for the comparison which should speed things up. Thoughts?

database/ticket.go
```diff
@@ -142,6 +144,13 @@ func (vdb *VspDatabase) DeleteTicket(ticket Ticket) error {
 }
 
 func (vdb *VspDatabase) UpdateTicket(ticket Ticket) error {
+	// Marshal ticket first. This could prevent a DB transaction being started
+	// and rolled back unnecessarily.
+	ticketBytes, err := json.Marshal(ticket)
```
dnldd, line 149: This would actually have the function do more work than needed if the ticket does not exist in the db. Having the typ marshalled on just before persisting is better I think.

database/votechange.go
```diff
@@ -27,13 +27,14 @@ type VoteChangeRecord struct {
 // one which is currently stored. Records are stored using a serially increasing
 // integer as the key.
 func (vdb *VspDatabase) SaveVoteChange(ticketHash string, record VoteChangeRecord) error {
+	// Marshal record first. This could prevent a DB transaction being started
+	// and rolled back unnecessarily.
+	recordBytes, err := json.Marshal(record)
```
dnldd, line 32: I agree with the marshal being outside of the db tx here, potentially prevents an rollback as intended.

jholdstock:
IMO, unnecessarily using a small amount of CPU/memory to marshal is a less expensive operation than unnecessarily starting/rolling back a database transaction.

In general, when designing systems, DB usage should be avoided/minimized wherever possible.

In practise, in the case of vspd, the likelihood of a marshal failure and the amount of DB access we are expecting are both so low that it probably doesn't make any difference. Even so, I'd like to keep these changes as they are, even if its just for the sake of setting a good example and following best practises.

database/ticket.go
jholdstock: The idea to change the DB keys would help to iterate over the tickets quicker, but would require a DB migration and is too much to change in this PR. We can investigate that idea later if the time to iterate becomes an issue in future.

jholdstock:
Discussed this with @dnldd on Matrix, and after researching and discovering that the chance for `json.Marshal` to return an error is practically zero, I have reverted most of the changes in this PR.

Keeping the PR open to fix some mtx.Lock/Unlock nits, and to defer marshalling a DB record until it is actually necessary.

dnldd (approved):

PR #215, opened 2020-12-22
Reject reused or old timestamps.

### LINKED ISSUE #14: timestamps in requests
Timestamps sent by clients cannot be trusted. Could potentally be useful if we need to process requests in some kind of order, but afaik this isn't necessary.

jholdstock:
Closes #14

### DIFF
--- a/webapi/errors.go
+++ b/webapi/errors.go
@@ -26,6 +26,7 @@ const (
 	errCannotBroadcastTicket
 	errCannotBroadcastFee
 	errCannotBroadcastFeeUnknownOutputs
+	errInvalidTimestamp
 )
 
 // httpStatus maps application error codes to HTTP status codes.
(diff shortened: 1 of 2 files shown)


webapi/errors.go
```diff
@@ -107,6 +110,8 @@ func (e apiError) defaultMessage() string {
 		return "fee transaction could not be broadcast"
 	case errCannotBroadcastFeeUnknownOutputs:
 		return "fee transaction could not be broadcast due to unknown outputs"
+	case errInvalidTimestamp:
+		return "request timestamp is not newer than all previous requests"
```
dnldd, line 114: s / request timestamp is not newer than all previous requests / request timestamp is older than the last request

jholdstock: Hm, your suggestion isnt quite accurate, but I take your point. I will update the message

dnldd (approved):

dajohi (approved):

PR #218, opened 2020-12-28
Make ticket parent mandatory

jholdstock:
Sending the parent hex param greatly reduces the chance of ticket broadcast errors. All clients have access to this information, and should be sending it, so there is no reason for this param to be optional.


jrick (approved):
sure

dajohi (approved):

PR #220, opened 2020-12-30
Use actual request in response, dont reserialize.

jholdstock:
Reserializing the client request can result in different bytes because
the order of iteration over a map is not guaranteed to be the same every
time.


dajohi (approved):

jrick:
This isn't correct.  It needs to use a json.RawMessage to avoid serializing the bytes of the request as a json array of numbers.

PR #226, opened 2021-01-15
Use json.RawMessage for echoed responses.

jholdstock:
Use json.RawMessage to avoid serializing the bytes of the request as a json array of numbers.


jholdstock:
Not needed

PR #239, opened 2021-04-20
Upgrade deps

jholdstock:
Includes using the dcrd stdaddr package instead of dcrutil.

Depends on #238


jholdstock:
Ready to go. @davecgh would you mind checking over the usage of `stdaddr`?

webapi/payfee.go
```diff
@@ -164,8 +164,8 @@ findAddress:
 		return
 	}
 
-	wifAddr, err := dcrutil.NewAddressPubKeyHash(dcrutil.Hash160(votingWIF.PubKey()), cfg.NetParams,
-		dcrec.STEcdsaSecp256k1)
+	pkHash := dcrutil.Hash160(votingWIF.PubKey())
```
davecgh, line 167: I'd suggest using `stdaddr.Hash160` here now.  This is fine for now, but the `dcrutil` version will likely eventually go away and since it's for the purposes of an address, it seems to make more sense to use `stdaddr` for it.

davecgh (commented):
The `stdaddr` usage looks correct.  I made one inline suggestion, but it's correct regardless.

davecgh (approved):

PR #243, opened 2021-05-05
Store each ticket in its own DB bucket.

### LINKED ISSUE #223: Improve ticket iterations & lookups
The criteria for ticket equality according to `InsertNewTicket` requires that the ticket hash as well as the fee address and the fee address index are identical. Tickets in the database are currently keyed by only their hashes however, requiring each ticket in the database being unmarshalled just for fee address and fee address index comparisons when checking for duplicates on inserting a new ticket into the database. I suggest using a composite key comprising of the fee address index, fee address and the ticket hash: `<fee address index><fee address><ticket hash>`. This composite key mitigates unmarshalling all tickets for comparison purposes when inserting a new ticket since all comparison data can be extracted from the key. 

If there are no objections I'd like to start making the necessary modifications and upgrade the database.

jholdstock:
**NOTE: This contains a backwards incompatible database migration, so if you plan to test it, please make a copy of your database first.**

Moves tickets from a single database bucket containing JSON encoded strings, to a bucket for each ticket.

This change is to preemptively deal with scaling issues seen with databases containing tens of thousands of tickets. (Closes #223)

### DIFF
--- a/database/ticket.go
+++ b/database/ticket.go
@@ -41,20 +41,20 @@ const (
 
 // The keys used to store ticket values in the database.
 var (
-	HashK              = []byte("Hash")
-	PurchaseHeightK    = []byte("PurchaseHeight")
-	CommitmentAddressK = []byte("CommitmentAddress")
-	FeeAddressIndexK   = []byte("FeeAddressIndex")
-	FeeAddressK        = []byte("FeeAddress")
-	FeeAmountK         = []byte("FeeAmount")
-	FeeExpirationK     = []byte("FeeExpiration")
-	ConfirmedK         = []byte("Confirmed")
-	VotingWIFK         = []byte("VotingWIF")
-	VoteChoicesK       = []byte("VoteChoices")
-	FeeTxHexK          = []byte("FeeTxHex")
-	FeeTxHashK         = []byte("FeeTxHash")
-	FeeTxStatusK       = []byte("FeeTxStatus")
-	OutcomeK           = []byte("Outcome")
+	hashK              = []byte("Hash")
+	purchaseHeightK    = []byte("PurchaseHeight")
+	commitmentAddressK = []byte("CommitmentAddress")
+	feeAddressIndexK   = []byte("FeeAddressIndex")
+	feeAddressK        = []byte("FeeAddress")
+	feeAmountK         = []byte("FeeAmount")
+	feeExpirationK     = []byte("FeeExpiration")
+	confirmedK         = []byte("Confirmed")
+	votingWIFK         = []byte("VotingWIF")
+	voteChoicesK       = []byte("VoteChoices")
+	feeTxHexK          = []byte("FeeTxHex")
+	feeTxHashK         = []byte("FeeTxHash")
+	feeTxStatusK       = []byte("FeeTxStatus")
+	outcomeK           = []byte("Outcome")
 )
 
 type Ticket struct {
@@ -113,7 +113,7 @@ func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
 		err = ticketBkt.ForEach(func(k, v []byte) error {
 			tbkt := ticketBkt.Bucket(k)
 
-			if string(tbkt.Get(FeeAddressK)) == ticket.FeeAddress {
+			if string(tbkt.Get(feeAddressK)) == ticket.FeeAddress {
 				return fmt.Errorf("ticket with fee address %s already exists", ticket.FeeAddress)
 			}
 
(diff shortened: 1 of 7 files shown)


database/ticket.go
```diff
@@ -114,23 +126,128 @@ func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
...
+	ticket.FeeTxHash = string(bkt.Get(FeeTxHashK))
+	ticket.FeeTxStatus = FeeStatus(bkt.Get(FeeTxStatusK))
+	ticket.Outcome = TicketOutcome(bkt.Get(OutcomeK))
+
+	ticket.PurchaseHeight = bytesToInt64(bkt.Get(PurchaseHeightK))
+	ticket.FeeAddressIndex = bytesToUint32(bkt.Get(FeeAddressIndexK))
+	ticket.FeeAmount = bytesToInt64(bkt.Get(FeeAmountK))
+	ticket.FeeExpiration = bytesToInt64(bkt.Get(FeeExpirationK))
+
+	// TODO is this dodgy?
```
jholdstock, line 226: Highlighting this particularly icky piece of code for reviewers - any better ideas for how to encode/decode `bool` & `[]byte`?

database/ticket.go
```diff
@@ -114,23 +126,128 @@ func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
...
+	if bkt.Get(ConfirmedK)[0] == byte(1) {
+		ticket.Confirmed = true
```
jholdstock, line 228: Highlighting this particularly icky piece of code for reviewers - any better ideas for how to encode/decode `bool` & `[]byte`?

database/ticket.go
```diff
@@ -114,23 +126,128 @@ func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
...
+	}
```
jholdstock, line 229: Highlighting this particularly icky piece of code for reviewers - any better ideas for how to encode/decode `bool` & `[]byte`?

jholdstock:
- Validated that key length makes no difference to database disk usage. Previously we used short keys because they were JSON encoded and contributed to increasing disk usage. Now we can use long, descriptive keys.
- Running on https://testnet-vsp.jholdstock.uk (database contains ~15k tickets). Memory usage and upgrade duration negligible.
- Used benchmark code from #223 to test a database containing ~200k tickets:
    - Old code
        - InsertNewTicket 2.39s
        - CountAllTickets 2.41s
    - New code
        - InsertNewTicket 0.30s
        - CountAllTickets 0.27s 

**Ready for review**

dnldd:
Still uneasy about the db size growth here, one thing I think we haven't looked at yet is using a custom encoder/decoder in the serialization/deserialization.We can borrow some ideas from how the block header is serialized/deserialized from dcrd. `readElement` and `writeElement` [here](https://github.com/decred/dcrd/blob/master/wire/common.go#L198) should be good references to put something together. Thoughts?

jholdstock:
While this kind of solution is useful for dcrd where messages need to be kept as tiny as possible for fast propagation through the network, I feel like that is a bit overkill for this case. It would add too much new complexity, whereas both the old solution and the new solution are relatively simple.

Interested to hear thoughts from @dhill as an experienced sysadmin. Would you mind reviewing comments on #223, particularly [this one](https://github.com/decred/vspd/issues/223#issuecomment-830177726)?

jholdstock:
I've spent a little more time on this. I've tried tweaking the bolt DB FillPercent parameters, I've tried reducing the number of buckets by encoding the vote choices as a JSON string rather than a bucket, I've tried using the internal compacting code from the bolt lib to try and make the db smaller. No progress.

In order to reduce filesize and keep the performance gains introduced by this PR, we either need to investigate a new encoding, as suggested by dnldd, or migrate to a new database

jholdstock:
One option which I think could be viable is to delete the raw fee transaction hex from the database once it has been broadcast and confirmed. That should reduce the space needed by each ticket quite significantly. I will run some tests.

dnldd:
I think removing the transaction hex is a good first step, that should get the db size down. Still think it'd be worthwhile getting a custom encoder/decoder done in the future borrowing from what's working in dcrd currently. That can come later if the confirmed tx hex are getting pruned from the db.

jholdstock:
Marking as a draft as this is now rebased and depends on #260

jholdstock:
- Benchmarking code:[bench_test.txt](https://github.com/decred/vspd/files/6514965/bench_test.txt)
- Full benchmarking results + graphs: [vspd_benchmarking.ods](https://github.com/decred/vspd/files/6514962/vspd_benchmarking.ods)

database/ticket.go
```diff
@@ -114,23 +127,114 @@ func (vdb *VspDatabase) InsertNewTicket(ticket Ticket) error {
...
-		return ticketBkt.Put(hashBytes, ticketBytes)
+		return nil
 	})
 }
 
-func (vdb *VspDatabase) DeleteTicket(ticket Ticket) error {
+func putTicketInBucket(bkt *bolt.Bucket, ticket *Ticket) error {
+	var err error
+	if err = bkt.Put(HashK, []byte(ticket.Hash)); err != nil {
+		return err
```
dnldd, line 142: Suggest wrapping these put errors to provide more context. A description for the function would be great too.

database/ticket.go
```diff
@@ -189,8 +288,7 @@ func (vdb *VspDatabase) GetTicketByHash(ticketHash string) (Ticket, bool, error)
 }
 
 // CountTickets returns the total number of voted, revoked, and currently voting
-// tickets. Requires deserializing every ticket in the db so should be used
-// sparingly.
+// tickets. Iterates over every ticket in the db so should be used sparingly.
```
dnldd, line 291: s / Iterates ... / This Iterates ... or It Iterates ...

webapi/admin.go
```diff
@@ -30,7 +30,7 @@ type WalletStatus struct {
 type searchResult struct {
 	Hash           string
 	Found          bool
-	Ticket         database.Ticket
+	Ticket         *database.Ticket
```
dnldd, line 33: Why the change to use a pointer here?

background/background.go
```diff
@@ -548,7 +548,7 @@ func checkWalletConsistency() {
 	}
 }
 
-func findOldestHeight(tickets []database.Ticket) int64 {
+func findOldestHeight(tickets []*database.Ticket) int64 {
```
dnldd, line 551: Why the change to use a slice of pointers here?

dnldd (changes requested):

webapi/admin.go
jholdstock: Whilst doing the work, there was a good reason to change this. However I think I made some later changes which made it unnecessary.

I have undone the change, no longer using pointers.

database/ticket.go
jholdstock: Updated

database/ticket.go
jholdstock: I've added a comment for the func.

I am reluctant to add additional error handling code here because it is already quite verbose/repetitive, and the chance of an error happening on these lines is negligible. Looking at the comments for `Put()`...

> Returns an error if the bucket was created from a read-only transaction,

Our code is always creating an Update transaction

> if the key is blank, if the key is too large,

Keys are hard-coded constant values which are neither blank nor too large (limit is 32768 chars)

> or if the value is too large.

Values are always of fixed or known length, and besides that, the maximum limit is 2,147,483,646

dnldd: gotcha, I initially wanted a more descriptive error considering its a number of clustered put calls, but this should be fine since in case of an error we can find the put call that erred with the line number.

dnldd (approved):

jholdstock:
Rebased onto master to pick up the latest linter.

xaur:
> Benchmarking code:[bench_test.txt](https://github.com/decred/vspd/files/6514965/bench_test.txt)

Would this code be useful in the repo?

Also curious why `InsertNewTicket` took 0.30 s in the [May 6](https://github.com/decred/vspd/pull/243#issuecomment-833436508) comment but 0.86 s in the [May 20](https://github.com/decred/vspd/pull/243#issuecomment-844945549) benchmark, while both are on a 200K ticket database. I would expect the latter would be even faster after #260.

jholdstock:
Adding the benchmark to the repo would likely be more of a burden than helpful.

- Its not useful to run in CI - it takes too long and the results cannot easily be interpreted automatically.
- If the code is not being run regularly, it will be liable to fall out of date with the rest of the codebase and need maintenance effort. Maintaining code which isn't being used is not fun.

As for the time differences, I carried out the testing on different machines, albeit with similar-ish specs. Also its likely that I was running other tasks at the same time as the benchmark which would have had an impact. Tbh the important point of the testing is not the exact timing for any given test, but the growth of the times. With this kind of database access we want to ensure that any growth is linear (or better) and not exponential.

PR #247, opened 2021-05-08
Add block explorer links

### LINKED ISSUE #245: Links to block explorer
When blocks, addresses, or transactions are referenced in the web UI, we can link to dcrdata to provide further information.

jholdstock:
- Add block explorer links for ticket hash, fee address and fee tx hash.
- Add a formatted date/time for fee expiration rather than only showing unix timestamp.

Closes #245

### DIFF
--- a/params.go
+++ b/params.go
@@ -33,5 +33,5 @@ var simNetParams = netParams{
 	Params:              chaincfg.SimNetParams(),
 	DcrdRPCServerPort:   "19556",
 	WalletRPCServerPort: "19557",
-	BlockExplorerURL:    "https://dcrdata.decred.org",
+	BlockExplorerURL:    "...",
 }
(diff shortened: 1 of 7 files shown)


params.go
```diff
@@ -12,22 +12,26 @@ type netParams struct {
...
 	DcrdRPCServerPort:   "19109",
 	WalletRPCServerPort: "19110",
+	BlockExplorerURL:    "https://testnet.dcrdata.org",
 }
 
 var simNetParams = netParams{
 	Params:              chaincfg.SimNetParams(),
 	DcrdRPCServerPort:   "19556",
 	WalletRPCServerPort: "19557",
+	BlockExplorerURL:    "https://dcrdata.decred.org",
```
dnldd, line 36: Since there isn't a simnet block explorer, a placeholder here would do? Ellipses?

jholdstock: Updated

dnldd (approved):

PR #255, opened 2021-05-14
Use a custom Recovery middleware.

### LINKED ISSUE #246: Log panic recovery
gin framework has a built-in middleware to recover from request handler panics, however it only logs these events to stdout. We should ensure that they are also added to log files.

jholdstock:
This has several advantages over the middleware provided by gin:

- It will log to the log file and not just stdout. (Closes #246)
- It will log the whole HTTP request which caused the panic (headers and body).
- Log message doesnt include any ANSI color codes, so log file is tidier.

The code to format the stack trace is mostly inherited from the gin recovery.go

### DIFF
--- a/webapi/recovery.go
+++ b/webapi/recovery.go
@@ -14,8 +14,10 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-// Recovery returns a middleware for a given writer that recovers from any
-// panics and calls the provided handle func to handle it.
+// Recovery returns a middleware that recovers from any panics which occur in
+// request handlers. It logs the panic, a stack trace, and the full request
+// details. It also ensure the client receives a 500 response rather than no
+// response at all.
 func Recovery() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		defer func() {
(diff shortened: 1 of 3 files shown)


webapi/recovery.go
```diff
@@ -0,0 +1,115 @@
...
+		return []byte("???")
+	}
+	name := []byte(fn.Name())
+	// The name includes the path name to the package, which is unnecessary
+	// since the file name is already included.  Plus, it has center dots.
+	// That is, we see
+	//	runtime/debug.*T·ptrmethod
+	// and want
+	//	*T.ptrmethod
+	// Also the package path might contains dot (e.g. code.google.com/...),
```
dnldd, line 105: s / contains / contain

dnldd (commented):
Looks good overall.

webapi/recovery.go
jholdstock: Fixed that, and also reviewed/updated some of the other comments.

webapi/recovery.go
```diff
@@ -0,0 +1,117 @@
...
+
+// Recovery returns a middleware that recovers from any panics which occur in
+// request handlers. It logs the panic, a stack trace, and the full request
+// details. It also ensure the client receives a 500 response rather than no
+// response at all.
+func Recovery() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		defer func() {
+			if err := recover(); err != nil {
+				// Check for a broken connection, as it is not really a
```
dnldd, line 25: Not an issue but you could avoid the additional indentation here if you checked `err == nil` and handled the error case after that.

dnldd (approved):

PR #257, opened 2021-05-17
Add timezone to GUI timestamps

jholdstock:


webapi/formatting.go
```diff
@@ -24,5 +24,5 @@ func blockURL(blockExplorerURL string) func(int64) string {
 }
 
 func dateTime(t int64) string {
-	return time.Unix(t, 0).Format("2 Jan 2006 15:04:05")
+	return time.Unix(t, 0).Format("2 Jan 2006 15:04:05 MST")
```
dnldd, line 27: Isn't UTC preferable here?

jholdstock: Just as writing 2006 is a template value to be replaced with the actual year, MST is a template value to be replaced with the local timezone.

On my server, UTC is the timezone, so UTC will be written. Servers running on other timezones will have different values.

dnldd (approved):

PR #260, opened 2021-05-18
Remove confirmed fee tx hex.

jholdstock:
This PR introduces a database migration which deletes any raw fee transactions which remain in the database after already having been confirmed on-chain. There is no need to keep these, and they take up a lot of space.

There is also a change in the background handler to ensure that in future, tx hex is always removed from the database when the tx is confirmed.

Before this PR, a 200k ticket database was ~370MB. After running the database migration, the database grows to 660MB, however that goes down to 126MB after running `bbolt compact`.

I will be sure to mention `bbolt compact` in the release notes in case any VSP operators wish to use it.

### DIFF
--- a/background/background.go
+++ b/background/background.go
@@ -180,6 +180,8 @@ func blockConnected() {
 		// If fee is confirmed, update the database and add ticket to voting
 		// wallets.
 		if feeTx.Confirmations >= requiredConfs {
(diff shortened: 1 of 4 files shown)


dnldd (approved):
Looks good :+1:

PR #264, opened 2021-06-01
Add network proportion to homepage and /vspinfo (and revoked proportion)

### LINKED ISSUE #206: show vsp proportion
The old vsp show us the vsp proportion of the pool and we should show it on this new vsp, as well:

image

jholdstock:
Proportion is calculated using the number of tickets  currently registered with the VSP, divided by the total size of the network ticket pool as reported by `getblockheader`.

The value will only ever be an estimate because:

- it's possible for a single ticket to be added to multiple VSPs.
- vspd does not distinguish between immature and live tickets, whereas `getblockheader` only reports live tickets.
- `getblockheader` is reporting the size of the ticket pool as of the previous block, not the current block.

Closes #206

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -55,7 +55,7 @@ when a VSP is closed will result in an error.
         "voted":25,
         "revoked":3,
         "blockheight":623212,
-        "networkproportion":0.04847841472045294
+        "estimatednetworkproportion":0.048478414
     }
     ```
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -54,7 +54,8 @@ when a VSP is closed will result in an error.
         "voting":10,
         "voted":25,
         "revoked":3,
-        "blockheight":623212
+        "blockheight":623212,
+        "networkproportion":0.04847841472045294
     }
     ```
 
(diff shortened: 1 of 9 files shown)


xaur:
If it will only ever be an estimate, how about reflecting that in the field name? e.g. `networkproportionestimate`. That makes the name long though :S

> `"networkproportion":0.04847841472045294`

I think 17 decimal places is an overkill. The lowest non-zero proportion is `1 / 50,000 = 0.00002`, assuming the ticket pool ever reaches 50K. Let's take a less rounded number, `1 / 49,998 ~= 0.000020001` - still only 9 places.

jholdstock:
I actually had all of those exact same thoughts myself. Eventually settled on using the shorter name rather than the more descriptive one, and used a `float64` rather than a shorter `float32` in the API simply to be more consistent with the other fields in the API.

I dont feel particularly strongly about either, and will happily change both to match your suggestions.

jholdstock:
Because this is a rather small PR and it hasnt received a technical review yet, I've included another commit which also displays the number of revoked tickets as a percentage

Screenshot from 2021-06-08 17-59-59

dajohi (approved):

PR #267, opened 2021-06-04
Compare scripts rather than addresses.

### LINKED ISSUE #266: Review usage of txscript vs stdscript
vspd uses `ExtractPkScriptAddrs` from the dcrd package `txscript` in two places

https://github.com/decred/vspd/blob/dad662b0f0e725c50f6aa095263f9ed3dccb71fc/webapi/payfee.go#L144-L151

https://github.com/decred/vspd/blob/dad662b0f0e725c50f6aa095263f9ed3dccb71fc/webapi/payfee.go#L186-L191

dcrd is adding a `stdscript` package which supersedes some elements of `txscript`. Once it is exposed, we should review whether vspd should be updated to use it.

Ref: decred/dcrd#2656

jholdstock:
Improve how voting WIF and fee payments are validated. Rather than just comparing address strings, compare scripts and script versions. Also fixes a bug where vspd would have rejected fee payment transactions which included unrelated outputs if they were not script version 0.

Removes usage of `txscript.ExtractPkScriptAddrs`, which in turn removes the need for hardcoded values `isTreasuryEnabled = true` and `scriptVersion = 0`.

Split into two commits to ease review.

Thanks for the suggestion @davecgh. Closes #266.

### DIFF
--- a/webapi/payfee.go
+++ b/webapi/payfee.go
@@ -6,22 +6,17 @@ package webapi
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
 	"time"
 
 	"github.com/decred/dcrd/blockchain/v4"
 	"github.com/decred/dcrd/dcrutil/v4"
-	"github.com/decred/dcrd/txscript/v4"
 	"github.com/decred/dcrd/txscript/v4/stdaddr"
 	"github.com/decred/vspd/database"
 	"github.com/decred/vspd/rpc"
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
-)
-
-const (
-	// Assume the treasury is enabled
-	isTreasuryEnabled = true
 )
 
 // payFee is the handler for "POST /api/v3/payfee".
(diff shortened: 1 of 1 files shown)


webapi/payfee.go
```diff
@@ -128,42 +124,47 @@ func payFee(c *gin.Context) {
...
-				feePaid = dcrutil.Amount(txOut.Value)
-				break findAddress
-			}
+		if txOut.Version == wantScriptVer && bytes.Equal(txOut.PkScript, wantScript) {
+			feePaid = dcrutil.Amount(txOut.Value)
+			break
 		}
 	}
 
+	// Confirm a fee payment was found.
```
davecgh, line 148: Unless you want to specifically distinguish the errors in the case of no payment vs paying too little, you could remove this check altogether since the next minimum fee check would catch it too, assuming `ticket.FeeAmount` is already prevented from being zero previously, of course.

davecgh (commented):
I would definitely recommend testing it out via testnet or simnet prior to release, but the logic reads correctly and is certainly more efficient than the previous.

I did point out another potential modification you might want to make inline depending on the desired result in terms of errors.  Otherwise, looks good.

jholdstock:
I've tried this out on testnet with the following scenarios:

- success with exact fee amount
- success with an oversized fee
- failure due to low fee payment to the correct address
- failure due to missing fee payment
- failure due to using WIF from incorrect ticket

webapi/payfee.go
jholdstock: There are different error codes for incorrect fee amount vs no fee, hence the seperate checks

davecgh (approved):

Issue #232, opened 2021-01-24
vspinfo endpoint ideas


xaur:
Some ideas from looking at `vspinfo` API [response](https://ultravsp.uk/api/v3/vspinfo).

Add fields:

- `statusmessage`: server operator's custom message with a sane character limit (must fit important messages + 1-2 links)
- `votingwalletsonline` and `votingwalletstotal` could be used by automated health check crawlers
- `ticketsmissed` (dcrstakepool reported this)

Change fields:

- `vspclosed`
  - first, the code often looks a bit cleaner if booleans have positive boolean names (`...up`, `...enabled`)
  - second, a string-based enum would capture more possible statuses than just up and down. I would call it just `status`

Group ticket stats in a namespace:

- rename `voting` to `ticketslive`
- rename `revoked` to `ticketsrevoked`
- rename `voted` to `ticketsvoted`

Related: https://github.com/decred/dcrstakepool/issues/628#issuecomment-628155661

jholdstock:
Cannot add missed tickets, see comment in #268

jholdstock:
Your renaming suggestions do make sense, but the existing names will be kept as they are because changing them will break existing API consumers for minimal benefit.

`vspclosed` is a boolean because it has exactly two states - either the VSP is currently accepting new tickets or it is not. Perhaps `acceptingtickets` would be a better name, but I refer back to the previous point.

There is already [an endpoint](https://github.com/decred/vspd/blob/master/docs/deployment.md#voting-wallet-status) for checking voting wallet status, but it requires authenticating with the admin credentials. We could explore opening this up for public consumption, but please open a new dedicated issue for that discussion because its a slightly wider topic than just updating the `/vspinfo` endpoint.

xaur:
Agreed re keeping the names to not break consumers. iirc I had this idea earlier but didn't submit it with good advance time before v1.0 final. Do you think it could be batched with other breaking changes, e.g. in v2.0, or will it still not worth it?

jholdstock:
Will definitely consider it if/when a new API version is introduced. Some possible scenarios which may make a new API version necessary:

- Improving the UX for staking on Trezor (see conversation [starting here](https://matrix.to/#/!MYLcxlwzxwViTaFPGo:decred.org/$lV5bgNkNUimlwYU2JzLZEs9YJ_gpGEoVNGJ7mG7PiWw?via=decred.org&via=matrix.org&via=planetdecred.org))
[...]

PR #269, opened 2021-06-08
Add vspclosedmsg config.

jholdstock:
If vspclosed is set to true, the provided message will be displayed on the webpage and returned by the status API endpoint.

Suggested by @xaur in #232

### DIFF
--- a/config.go
+++ b/config.go
@@ -62,6 +62,7 @@ type config struct {
 	SupportEmail    string        `long:"supportemail" ini-name:"supportemail" description:"Email address for users in need of support."`
 	BackupInterval  time.Duration `long:"backupinterval" ini-name:"backupinterval" description:"Time period between automatic database backups. Valid time units are {s,m,h}. Minimum 30 seconds."`
 	VspClosed       bool          `long:"vspclosed" ini-name:"vspclosed" description:"Closed prevents the VSP from accepting new tickets."`
+	VspClosedMsg    string        `long:"vspclosedmsg" ini-name:"vspclosedmsg" description:"A short message displayed on the webpage and returned by the status API endpoint if vspdclosed is true."`
 	AdminPass       string        `long:"adminpass" ini-name:"adminpass" description:"Password for accessing admin page."`
 	Designation     string        `long:"designation" ini-name:"designation" description:"Short name for the VSP. Customizes the logo in the top toolbar."`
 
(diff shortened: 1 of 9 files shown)


config.go
```diff
@@ -292,6 +293,11 @@ func loadConfig() (*config, error) {
 		return nil, errors.New("invalid vspfee - should be greater than 0.01 and less than 100.0")
 	}
 
+	// If VSP is not closed, ignore any provided closure message.
+	if !cfg.VspClosed {
+		cfg.VspClosedMsg = ""
```
dnldd, line 298: Not sure why this needs to be set to nil if there is a flag available which can be checked before the message is used.

webapi/templates/homepage.html
```diff
@@ -5,9 +5,16 @@
...
         {{ if .VspStats.VspClosed }}
             <div class="alert alert-danger">
-                <h4 class="alert-heading">This Voting Service Provider is closed</h4>
-                A closed VSP will still vote on tickets with already paid fees, but will not accept new any tickets.
-                Visit <a href="https://decred.org/vsp/" class="alert-link" target="_blank" rel="noopener noreferrer">decred.org</a> to find a new VSP.
+                <h4 class="alert-heading mb-3">
+                    This Voting Service Provider is closed
+                </h4>
+                <p>
+                    {{ .VspStats.VspClosedMsg }}
```
dnldd, line 12: Consider checking the VspClosed flag before displaying the message. I think that's better than setting the config to an empty string.

dnldd (changes requested):

webapi/templates/homepage.html
jholdstock: Its already done

config.go
jholdstock: This same `if` statement will still need to exist in `vspinfo.go` to ensure that the message is only included in `/vspinfo` response when the VSP is actually closed.

I put it in `config.go` instead because the value of `VspClosedMsg` should never be used when `VspClosed == true`, so putting it here will prevent the need for it to be duplicated in other places which might access the closed message in future.

webapi/templates/homepage.html
dnldd: Ah my bad, just noticed.

config.go
dnldd: I get that, but ignoring the closure message by setting it to an empty string isn't intuitive to me, if the checking the flag precedes using the closure message then it'd only be displayed if the check passes anyway so why mutate the config value? Plus the check isn't expensive. The current approach works and I do not have a strong opinion on this, take another look however.

dnldd (approved):

PR #271, opened 2021-06-09
Add shutdownWg to dcrd notif handler.

jholdstock:
This prevents db/dcrd/dcrwallet connections from being closed while the dcrd notification handler is still running.

Marking this as a draft because I want to run it for a while before merging it.


jholdstock:
Ready to review. I've been running this in local dev, and on my server for a few days and it seems to work well.

dnldd (approved):

Issue #234, opened 2021-01-26
Implement tspend policy operations APIs


matheusd:
Policy choice for tspends can be set on dcrwallet for both a specific pi key (that is, approving or disapproving all tspends signed by some given key) or by individual tspend hash.

This needs to be exposed so that stakeholders can select their preference for when their vote is cast.

At a minimum, users need to be able to set and retrieve the vote choices for a given ticket.

jholdstock:
### Add optional params to `/payFee` request

The client is currently setting their consensus vote choices when a ticket is initially registered with the `/payFee` request. We should add new, optional params to also set tspendPolicy and piKeyPolicy. eg.

```
...
"tspendPolicy": { "first_tspend_hash": "yes", "second_tspend_hash": "no" },
"piKeyPolicy": { "first_pi_key": "key", "second_pi_key": "key" },
...
```

### New endpoints for updating tspendPolicy/piKeyPolicy

- `/api/v3/settspendpolicy`
- `/api/v3/setpikeypolicy`

Using the same request encoding specified above, and the existing mechanisms for ticket authentication and storing request history for accountability.

### Additional thoughts

- decred/dcrwallet#2051 needed first
- Valid values for vote choices are "invalid", "abstain", "yes", "no". Reject anything else.
- Ensure tspend/pikey policy are being set in all places where consensus vote choices are currently being set, ie:
    - Voting wallet consistency checking
    - In the background handler, when a ticket fee tx is confirmed
- We don't need to bump API version here because all new fields/endpoints are optional.

davecgh:
The mostly seems fine, but I don't see a way to disallow a specific pi key (for example, if it gets compromised, etc).  Simply not specifying one would result in an abstain as opposed to explicitly voting no.

jholdstock:
Typo - I fixed the second `piKeyPolicy` example above

jholdstock:
I've started implementing this and rather than creating new endpoints for `/setspendpolicy` and `/settreasurypolicy`,  I think it makes more sense to add new optional fields to the existing `/setvotechoices` endpoint.

The name `setvotechoices` is already generic enough to cover both consensus voting and treasury spend voting. Internally to vspd, it will ease the process of recording vote choice changes. From a wider perspective, it will reduce the amount of network calls required to update choices, which in turn may reduce the amount of times accounts need to be unlocked in dcrwallet to sign requests.

jholdstock:
[...]

PR #278, opened 2021-06-21
Be more specific about consensus vote choices.

jholdstock:
Rather than referring to just "Vote choices", refer specifically to "consensus vote choices".

This will be useful when treasury and tspend vote choices are added, and the distinction becomes more important.

This is similar to #262 - minor standalone bits which were changing while implementing #234

### DIFF
--- a/webapi/setvotechoices.go
+++ b/webapi/setvotechoices.go
@@ -120,7 +120,7 @@ func setVoteChoices(c *gin.Context) {
 		// to attempt to update as much as possible regardless of any errors.
 		for _, walletClient := range walletClients {
 
-			// Consensus vote choices.
+			// Set consensus vote choices.
 			for agenda, choice := range request.VoteChoices {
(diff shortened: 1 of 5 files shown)


webapi/setvotechoices.go
```diff
@@ -115,16 +115,20 @@ func setVoteChoices(c *gin.Context) {
 	// Update vote choices on voting wallets. Tickets are only added to voting
 	// wallets if their fee is confirmed.
 	if ticket.FeeTxStatus == database.FeeConfirmed {
+
+		// Just log any errors which occur while setting vote choices. We want
+		// to attempt to update as much as possible regardless of any errors.
 		for _, walletClient := range walletClients {
-			for agenda, choice := range voteChoices {
+
+			// Consensus vote choices.
```
dnldd, line 123: s/  Consensus vote choices. /  Set consensus vote choices.

dnldd (approved):

PR #282, opened 2021-06-24
cache: Avoid unlikely zero division.

JoeGruffins:


jholdstock (commented):
I actually considered adding this, but didn't do it because if `bestBlock.PoolSize == 0` probably indicates some kind of underlying issue - eg. misconfigured or misbehaving dcrd. I wanted the panic to occur so it would be logged, and could be investigated.

HOWEVER. At that time, I thought this code was inside of a web request handler, so it would be covered by the panic recovery offered by `recovery.go`. I've just realised that this is not the case, and actually a panic in this code would bring down the whole vspd process.

 Perhaps the best option here is to return an error without updating the cached stats. That will ensure that we are still displaying accurate (if outdated) info on the webpage, and also ensures the unexpected event will show up in the logs.

Thoughts?

JoeGruffins:
I thought it might be something that comes up using the harness sometimes.

Sure, returning an error is maybe ok. It could also be a problem in dcrd that randomly causes this, rather than a problem with vspd.

I'll change this to error if we get 0 for the ticket pool.

dnldd:
Update seems fine but wondering if logging there aren't any tickets to cache and returning wouldn't be the better option here.

JoeGruffins:
> Update seems fine but wondering if logging there aren't any tickets to cache and returning wouldn't be the better option here.

Do you mean not error?

dnldd:
> Do you mean not error?

Yeah.

JoeGruffins:
I thought it was probably fine not to error. It's super unlikely to happen in the first place. I can only imagine something wrong with the rcp return. The chain doesn't move without votes. @jholdstock log and continue or return error?

dajohi (approved):

jholdstock (approved):

PR #287, opened 2021-06-28
mulit: Add setaltsig endpoint.

### LINKED ISSUE #279: api: Add `addaltsig` endpoint.
Problem:

Trezor cannot passively sign messages to vspd. This presents some difficulties for managing tickets purchased with trezor.

Solution:

Allow adding an optional alternate signing path via `POST /api/v3/addaltsig`. This endpoint will require having an active ticket, and a signature corresponding to that ticket's commitment address. It accepts a public key that can be used in future requests to verify the signature for requsting the fee, paying the fee, retrieving ticket status, and changing vote preferences. Only one alternate need be saved, and calling this method again will set the current alt signature`s address to the new one without error. The address that verifies the alternate signature will be saved in persistent storage.

Preexisting endpoints will need an optional boolean field called `usealtsig` which is false by default. If true, signatures of json message will be verified using the address in storage. I believe that using a separate route to add an alt sig means this is backwards compatible with current vspd. If the `.../addaltsig` doesn't exist on the vspd, the caller can assume the vspd is the wrong version and fail before attempting to request the fee (but after buying the ticket...).

JoeGruffins:
closes #279


database/database_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 The Decred developers
+// Copyright (c) 2021 The Decred developers
```
JoeGruffins, line 1: whoops, should be 2020-2021

database/ticket.go
```diff
@@ -244,6 +291,84 @@ func (vdb *VspDatabase) UpdateTicket(ticket Ticket) error {
 	})
 }
 
+// AltSigHistory retrieves the history of changes to a ticket's alternate signature.
+func (vdb *VspDatabase) AltSigHistory(ticketHash string) ([]*AltSigHistory, error) {
+	vdb.ticketsMtx.RLock()
+	defer vdb.ticketsMtx.RUnlock()
+
+	history := make([]*AltSigHistory, 0)
```
dnldd, line 299: Wouldn't this be capped to `maxAltSigs`? If that's the case then pre-allocating would be ideal, `history := make([]*AltSigHistory, 0, maxAltSigs)`

database/ticket.go
```diff
@@ -244,6 +291,84 @@ func (vdb *VspDatabase) UpdateTicket(ticket Ticket) error {
...
+// error. These records should never be deleted or altered as the server needs
+// them to prove the client's current alternate signature was indeed chosen by
+// the client.
+func (vdb *VspDatabase) AddAltSigHistory(ticketHash string, history *AltSigHistory) error {
+	vdb.ticketsMtx.Lock()
+	defer vdb.ticketsMtx.Unlock()
+	return vdb.db.Update(func(tx *bolt.Tx) error {
+		ticketBkt := tx.Bucket(vspBktK).Bucket(ticketBktK).Bucket([]byte(ticketHash))
+		if ticketBkt == nil {
+			return fmt.Errorf("not bucket for ticket %v", ticketHash)
```
dnldd, line 336: s / not bucket for ticket / no bucket for ticket.

database/ticket_test.go
```diff
@@ -67,14 +78,6 @@ func testInsertNewTicket(t *testing.T) {
...
 	}
 
-	// Inserting a ticket with the same fee address should fail.
-	ticket3 := exampleTicket()
-	ticket3.FeeAddress = ticket.FeeAddress
-	err = db.InsertNewTicket(ticket3)
-	if err == nil {
-		t.Fatal("expected an error inserting ticket with duplicate fee addr")
-	}
-
```
dnldd, line 77: Not sure why this was removed, `testHasFeeAddress` isn't checking for persisting a ticket with duplicate a duplicate fee address.

database/ticket_test.go
```diff
@@ -83,6 +86,32 @@ func testInsertNewTicket(t *testing.T) {
...
+	has, err = db.HasFeeAddress(ticket.FeeAddress)
+	if err != nil {
+		t.Fatalf("error retrieving has fee address: %v", err)
+	}
+	if !has {
+		t.Fatal("unexpected has for fee address")
+	}
+
+}
+
```
dnldd, line 114: I think the error messages could be more descriptive here, without looking at the code `has` in the error messages has very little context. For example,`unexpected has for fee address` could be `expected db to have ticket with fee address <addr>`.

database/ticket_test.go
```diff
@@ -332,3 +362,110 @@ func testCountTickets(t *testing.T) {
...
+	// This should error if no ticket exists for hash in the database.
+	if err := db.AddAltSigHistory(ticket.Hash, tAltHist()); err == nil {
+		t.Fatal("expected error for no ticket in db")
+	}
+
+	// Insert a ticket into the database.
+	err := db.InsertNewTicket(ticket)
+	if err != nil {
+		t.Fatalf("error storing ticket in database: %v", err)
```
dnldd, line 385: s / error storing ticket in database / unexpected error storing ticket in database.

database/ticket_test.go
```diff
@@ -332,3 +362,110 @@ func testCountTickets(t *testing.T) {
...
+	if err := db.AddAltSigHistory(ticket.Hash, sameHist); err != nil {
+		t.Fatalf("unexpected error for third alt history: %v", err)
+	}
+
+	// Over the max, currently 3, should error.
+	if err := db.AddAltSigHistory(ticket.Hash, sameHist); err == nil {
+		t.Fatalf("expected error for fourth alt history: %v", err)
+	}
+	if err := db.AddAltSigHistory(ticket.Hash, tAltHist()); err == nil {
+		t.Fatalf("expected error for second fourth alt history: %v", err)
```
dnldd, line 407: s / ... second fourth alt history / ... fifth alt history

database/ticket_test.go
```diff
@@ -332,3 +362,110 @@ func testCountTickets(t *testing.T) {
...
+			t.Fatalf("expected history len of %d but got %d", i+1, len(history))
+		}
+		if !reflect.DeepEqual(wantHist, history) {
+			t.Fatalf("expected history %v but got %v", spew.Sdump(wantHist), spew.Sdump(history))
+		}
+	}
+
+	// Further additions should error and not change the history.
+	if err := db.AddAltSigHistory(ticket.Hash, tAltHist()); err == nil {
+		t.Fatalf("expected error for forth alt history addition")
```
dnldd, line 461: s / ...forth / ...fourth

database/ticket_test.go
```diff
@@ -332,3 +362,110 @@ func testCountTickets(t *testing.T) {
...
+	}
+
+	history, err = db.AltSigHistory(ticket.Hash)
+	if err != nil {
+		t.Fatalf("unexpected error for fourth alt history: %v", err)
+	}
+	if !reflect.DeepEqual(wantHist, history) {
+		t.Fatalf("expected history %v but got %v", spew.Sdump(wantHist), spew.Sdump(history))
```
dnldd, line 469: Should probably count the history entries here and drop spew as a dependency.

dnldd (changes requested):
Yet to finish reviewing everything, first batch of suggestions here.

database/ticket_test.go
JoeGruffins: I removed it from that function, because if this endpoint goes in, it will create a new path that does not need to check the fee address, because it doesn't put one, and then if it's already in the db but with no fee address, this will need to be checked when the first fee address goes in, so outside of insert, during an update.

This also brings up the question, why aren't we checking **every** fee address that is created? imo the check isn't needed at all, but why only on inserts? A new address can also be added during update in master now. It seems the "risk" of inserting a duplicate address is the same for either insert or update.

database/ticket_test.go
JoeGruffins: aw....

database/ticket.go
JoeGruffins: ok

database/ticket.go
JoeGruffins: ok

database/ticket_test.go
JoeGruffins: ok

database/ticket_test.go
JoeGruffins: hmmmm

JoeGruffins: ok

database/ticket_test.go
JoeGruffins: ok

database/ticket_test.go
JoeGruffins: ok

database/ticket_test.go
JoeGruffins: ok

JoeGruffins:
As discussed in matrix, Will change the history to one record.

database/upgrades.go
```diff
@@ -55,6 +55,9 @@ type v1Ticket struct {
 	FeeTxHash         string            `json:"fhsh"`
 	FeeTxStatus       FeeStatus         `json:"fsts"`
 	Outcome           TicketOutcome     `json:"otcme"`
+	// AltSigAddress was added later. Here to allow converting to the new
+	// type.
+	AltSigAddress string
```
jholdstock, line 60: Adding this field to `v1Ticket` means it is no longer representative of a real version 1 ticket. We can avoid adding this field if we change the conversion in `upgrade_v3.go` to be explicit, ie:

```
newTicket := Ticket{
    Hash: v1Ticket.Hash,
    ...
}

```

This will also benefit my upcoming PR which adds consensus vote choices to `Ticket`, and any other future work which adds new fields.

JoeGruffins: Did this I think. 6e98b01

database/ticket.go
```diff
@@ -244,6 +291,84 @@ func (vdb *VspDatabase) UpdateTicket(ticket Ticket) error {
...
+		}
+		return nil
+	})
+}
+
+// AddAltSigHistory adds an alternate signature record to history. If no bucket
+// exists for this ticket an error is returned. If the altSigHistory bucket does
+// not yet exist for the ticket, one is created. Records are keyed by number of
+// records plus one. Only maxAltSigs are allowed, and trying to add more returns
+// error. These records should never be deleted or altered as the server needs
```
dnldd, line 327: s / returns error / returns an error.

database/ticket_test.go
```diff
@@ -332,3 +361,101 @@ func testCountTickets(t *testing.T) {
...
+	}
+}
+
+func testAltSigHistory(t *testing.T) {
+	ticket := exampleTicket()
+
+	// No ticket means no history.
+	history, err := db.AltSigHistory(ticket.Hash)
+	if err != nil {
+		t.Fatalf("unexpected error for no ticket in db: %v", err)
```
dnldd, line 407: s / unexpected error for no ticket in db / unexpected error fetching alt signature history for ticket

database/ticket_test.go
```diff
@@ -332,3 +361,101 @@ func testCountTickets(t *testing.T) {
...
+	// Insert a ticket into the database.
+	err = db.InsertNewTicket(ticket)
+	if err != nil {
+		t.Fatalf("unexpected error storing ticket in database: %v", err)
+	}
+
+	// No history added yet.
+	history, err = db.AltSigHistory(ticket.Hash)
+	if err != nil {
+		t.Fatalf("unexpected error for new: %v", err)
```
dnldd, line 422: This error seems incomplete and should be about the alt signature history.

database/ticket_test.go
```diff
@@ -332,3 +361,101 @@ func testCountTickets(t *testing.T) {
...
+		if err := db.AddAltSigHistory(ticket.Hash, altHist); err != nil {
+			t.Fatalf("unexpected error for first alt history: %v", err)
+		}
+
+		history, err = db.AltSigHistory(ticket.Hash)
+		if err != nil {
+			t.Fatalf("unexpected error for alt history %d: %v", i+1, err)
+		}
+		if len(history) != i+1 {
+			t.Fatalf("expected history len of %d but got %d", i+1, len(history))
```
dnldd, line 442: s/ history len / history length

docs/api.md
```diff
@@ -65,6 +65,40 @@ when a VSP is closed will result in an error.
...
 its voting wallets unless both of these calls have succeeded.**
 
+#### Step Zero (optional)
+
+Set an alternate signing address for a ticket. The ticket must be valid and will
+be transmitted to the network if not found. If set, the vsp will check that a
+signature is good for this address and fallback to the commitment address if not.
+The address must be valid for the network and a pay to secp256k1 ecdsa pubkey
+hash script. Can be cleared by sending `""` as the `altpubaddress`. The key may
+be set a maximum of three times. Further requests to set a new key will be
```
dnldd, line 75: s / `The key may be set a maximum of three times.` / `The key currently can only be set once, users are advised to crosscheck inputs as many times as necessary.`

webapi/middleware.go
```diff
@@ -305,58 +305,77 @@ func vspAuth() gin.HandlerFunc {
...
-		var commitmentAddress string
-		if ticketFound {
-			commitmentAddress = ticket.CommitmentAddress
-		} else {
-			dcrdClient := c.MustGet("DcrdClient").(*rpc.DcrdRPC)
-
-			resp, err := dcrdClient.GetRawTransaction(hash)
+		// validate validates the signature in the request as signing
+		// the request body with private key corresponding to sAddr.
+		// fatal will cause the client to be notified of failure if
```
dnldd, line 310: s / fatal will cause ... / Verification failure results in the client getting notified of the cause.

s / Returns wether verification succeeded and sets some header values if success. / Returns whether verification succeeded.

webapi/setaltsig_test.go
```diff
@@ -0,0 +1,276 @@
...
+		if test.setMaxHistory {
+			for i := 0; i < 1; i++ {
+				request()
+			}
+		}
+		request()
+
+		resp := w.Result()
+		body, _ := ioutil.ReadAll(resp.Body)
+		fmt.Printf("\nThe response body for %q: %s\n\n", test.name, string(body))
```
dnldd, line 271: Debug line slipped through.

dnldd (changes requested):
Looks good overall, minor typos and errors to clean up.

webapi/setaltsig_test.go
JoeGruffins: I thought this was alright to leave in tests, as you can see the error going to the user, but I'll remove.

database/ticket.go
JoeGruffins: ok

database/ticket_test.go
JoeGruffins: ok

database/ticket_test.go
JoeGruffins: Made to be the same as the previous error log.

database/ticket_test.go
JoeGruffins: ok

webapi/middleware.go
JoeGruffins: ok

docs/api.md
JoeGruffins: ok

JoeGruffins:
This code is running at dcrstakedinner.com

database/ticket.go
```diff
@@ -55,8 +67,27 @@ var (
...
 
+// AltSigHistory holds the information needed to prove that a client changed
+// their alternate signature address.
+type AltSigHistory struct {
+	// Addr is the address used to validate Sig.
+	Addr string
+	// Req is the original request to set an alternate signature.
+	Req []byte
+	// Sig is the request's signature signed by the private key the
+	// corresponds to Addr.
```
dnldd, line 85:
```suggestion
	// Sig is the request's signature signed by the private key that
	// corresponds to the address.
```

dnldd (approved):
Looks good overall.

database/ticket.go
JoeGruffins: done

JoeGruffins:
Will go ahead and squash the current squashables tomorrow.

jholdstock:
I've been chatting with @JoeGruffins about this on Matrix. I'm not very happy with how much code this PR is currently changing/adding. This is a minor change to functionality, which is to support a small set of users, and I think we can achieve the desired functionality with a much smaller changeset, without altering too much of the existing code.

A new database bucket to store altSig, request and response (keyed with ticket hash) can be introduced. In the new `/altsig` endpoint, we just create an entry in that bucket. Then we update the auth middleware to use any existing values in that bucket, or fallback to using the commitment address.

I believe that would avoid all of the changes which are being made to `/getfeeaddress` and the existing database code. Further, isolating the new optional step in this way means that code which is executed later doesnt need to look like `if optional step completed (do this) else (do that)` - it can just stay as it currently is.

JoeGruffins:
Let me know if this looks more like what you were thinking @jholdstock  It does not touch getfeeaddress any longer, or mess with ticket initialization logic.

I do think that maybe we should remove addresses from the database for tickets that end up not being handled, like what happens with.. tickets I think.

This comes with a database upgrade now, for a new bucket. If you test please back up your original database first.

database/upgrades.go
```diff
@@ -25,17 +25,22 @@ const (
 	// moves each ticket into its own bucket and does away with JSON encoding.
 	ticketBucketVersion = 3
 
+	// ticketBucketVersion adds a bucket to store alternate signatures used to
```
jholdstock, line 28:
```suggestion
	// altSigVersion adds a bucket to store alternate signatures used to
```

database/upgrade_v4.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// Run the upgrade in a single database transaction so it can be safely
+	// rolled back if an error is encountered.
+	err := db.Update(func(tx *bolt.Tx) error {
+		vspBkt := tx.Bucket(vspBktK)
+
+		// Create altsig bucket.
+		_, err := vspBkt.CreateBucket(altSigBktK)
+		if err != nil {
+			return fmt.Errorf("failed to create %s bucket: %w", string(altSigBktK), err)
```
jholdstock, line 24: `string()` conversion here is unnecessary: https://play.golang.org/p/QZ99v4sRpoW

database/altsig.go
```diff
@@ -0,0 +1,208 @@
...
+	// Sig is the request's signature signed by the private key that
+	// corresponds to the address.
+	Sig []byte
+}
+
+// InsertAltSig will insert the provided ticket into the database. Returns an
+// error if either the ticket hash or fee address already exist.
+//
+// history must be included.
+func (vdb *VspDatabase) InsertAltSig(txid, altSigAddr string, history *AltSigHistory) error {
```
jholdstock, line 51:
```suggestion
func (vdb *VspDatabase) InsertAltSig(ticketHash, altSigAddr string, history *AltSigHistory) error {
```

I know its more verbose, but for the sake of maintaining consistency with the rest of the codebase, would like to stick with the more descriptive `ticketHash` here please.

Same for all other DB funcs.

webapi/setaltsig.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	var request SetAltSigRequest
+	if err := binding.JSON.BindBody(reqBytes, &request); err != nil {
+		log.Tracef("%s: Bad request (clientIP=%s): %v", funcName, c.ClientIP(), err)
+		sendErrorWithMsg(err.Error(), errBadRequest, c)
+		return
+	}
+
+	altPubAddr, txid := request.AltPubAddress, request.TicketHash
```
jholdstock, line 52:
```suggestion
	altPubAddr, ticketHash := request.AltPubAddress, request.TicketHash
```

jholdstock (changes requested):
Partial review added, and jotting down some thoughts:

- If a client adds an alt sig for a ticket, and that ticket never gets mined, `DeleteAltSig` should be called here: https://github.com/decred/vspd/blob/master/background/background.go#L91-L94

- Thus far we dont have an easy solution for deleting alt sigs for tickets which get mined but never get a fee payment. We can leave that for a later date (a new issue).

database/upgrade_v4.go
JoeGruffins: Thank you. removed conversion.

database/altsig.go
JoeGruffins: Ok.

webapi/setaltsig.go
JoeGruffins: Ok.

database/upgrades.go
JoeGruffins: whoops

JoeGruffins:
Deleting the altsig for unmined tickets. Also completely removed a notion of max altsigs and updating. Only allow users to set once.

This may not be optimal, but in order to use the new endpoint with decrediton, I needed to set some values in the nginx file. This is mine ver-batim.
<details closed>
<summary>nginx.conf</summary>

```
user nginx;
worker_processes auto;
error_log /var/log/nginx/stakepool_error.log;
pid /run/nginx.pid;

events {
    worker_connections 1024;
}

http {
    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for" "$realip_remote_addr"';

    access_log  /var/log/nginx/stakepool_access.log  main;

    sendfile            on;
    tcp_nopush          on;
    tcp_nodelay         on;
    keepalive_timeout   65;
    types_hash_max_size 2048;

    include             /etc/nginx/mime.types;
    default_type        application/octet-stream;

    # Ensure ngx_http_realip_module is available
    set_real_ip_from 127.0.0.1/32;
    # If you are behind a NAT router, specify LAN
    #set_real_ip_from 10.24.0/16;
    real_ip_header X-Forwarded-For;
    real_ip_recursive on;

    server {
        listen          80 default_server;
        server_name     _;

        return 308      https://$host$request_uri;
    }

    limit_req_zone $binary_remote_addr zone=stakepool:10m rate=8r/s;

    server {
        listen       443 http2 ssl default_server;
        server_name dcrstakedinner.com; # managed by Certbot

        gzip on;
        gzip_proxied any;
        gzip_comp_level 3;
        gzip_min_length 512;
                gzip_types text/css text/* text/javascript application/x-javascript application/json
                application/xml application/atom+xml application/xaml+xml application/javascript
                application/x-font-ttf application/font-woff font/collection font/opentype
                font/otf font/ttf image/bmp image/svg+xml image/x-icon application/octet-stream;

        #ssl                     on; # only needed for older nginx where not in listen
        ssl_certificate /etc/letsencrypt/live/dcrstakedinner.com/fullchain.pem; # managed by Certbot
        ssl_certificate_key /etc/letsencrypt/live/dcrstakedinner.com/privkey.pem; # managed by Certbot
 
        ssl_session_cache               shared:SSL:20m;
        ssl_protocols                   TLSv1.1 TLSv1.2; # add TLSv1.3 if supported
        ssl_ciphers                     EECDH+CHACHA20:EECDH+CHACHA20-draft:EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:!MD5;
        ssl_prefer_server_ciphers       on;
        #ssl_dhparam                     /etc/nginx/dhparam.pem; # openssl dhparam -out /etc/nginx/dhparam.pem 4096
        add_header                      Strict-Transport-Security max-age=15552001;



        # Everything else is proxied to dcrstakepool.
        location / {
            # apply rate limiting
            limit_req zone=stakepool burst=16;

            proxy_set_header Host $host;
            proxy_set_header X-Real-IP $realip_remote_addr;
            proxy_pass http://127.0.0.1:8000;
            proxy_http_version 1.0;
            if ($request_method = 'OPTIONS') {
                add_header 'Access-Control-Allow-Origin' "https://dcrstakedinner.com" always;
                add_header 'Access-Control-Allow-Credentials' 'true' always;
                add_header 'Access-Control-Allow-Headers' "Origin, X-Requested-With, Content-Type, Accept, vsp-client-signature" always;
                add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
                add_header 'Access-Control-Max-Age' 1728000;
                add_header 'Content-Type' 'text/plain; charset=utf-8';
                add_header 'Content-Length' 0;
                return 204;
            }
        }
    }
}
```
</details>

decrediton can be on this pr: https://github.com/decred/decrediton/pull/3533
dcrwallet for the client is currently the wrong rpcversion to work with decrediton master, so a temporary wallet commit is here: https://github.com/JoeGruffins/dcrwallet/tree/useaccoutforvoting.flat

The vspd server can be run with default dcrd/dcrwallet.

webapi/setaltsig.go
```diff
@@ -0,0 +1,137 @@
...
+	sigStr := c.GetHeader("VSP-Client-Signature")
+	sig, err := base64.StdEncoding.DecodeString(sigStr)
+	if err != nil {
+		log.Tracef("%s: error decoding signature string (ticketHash=%s, sig=%s): %v", funcName, ticketHash, sigStr, err)
+		sendError(errInternalError, c)
+		return
+	}
+	history := &database.AltSigHistory{
+		Addr: sigAddress,
+		Req:  reqBytes,
```
jholdstock, line 119: Thoughts on recording response and response signature here, as we do for vote change records?

database/altsig.go
```diff
@@ -0,0 +1,126 @@
...
+// AltSigHistory holds the information needed to prove that a client changed
+// their alternate signature address.
+type AltSigHistory struct {
+	// Addr is the address used to validate Sig.
+	Addr string
+	// Req is the original request to set an alternate signature.
+	Req []byte
+	// Sig is the request's signature signed by the private key that
+	// corresponds to the address.
+	Sig []byte
```
jholdstock, line 35:
```suggestion
	ReqSig []byte
```

Lets be specific about this because the name `sig` is used a lot in this code.

webapi/middleware.go
```diff
@@ -305,58 +305,93 @@ func vspAuth() gin.HandlerFunc {
...
-				log.Warnf("%s: Invalid ticket (clientIP=%s, ticketHash=%s): %v", funcName, c.ClientIP(), hash, err)
-				sendError(errInvalidTicket, c)
-				return
-			}
+			// Add ticket information to context so downstream handlers don't need
+			// to access the db for it.
+			c.Set("Ticket", *t)
+			c.Set("KnownTicket", kt)
+			c.Set("CommitmentAddress", t.CommitmentAddress)
+			c.Set("SigningAddress", sAddr)
```
jholdstock, line 352: I dont understand why this is being added into the context. It's only used in `setaltsig.go`, and I think the code there can just extract it from `SetAltSigRequest`.

If we are able to remove `aAddr`, that simplifies the code above a bit as well.

webapi/setaltsig.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if !canVote {
+		log.Tracef("%s: unvotable ticket (clientIP=%s, ticketHash=%s)",
+			funcName, c.ClientIP(), ticketHash)
+		sendError(errTicketCannotVote, c)
+		return
+	}
+
+	sigStr := c.GetHeader("VSP-Client-Signature")
+	sig, err := base64.StdEncoding.DecodeString(sigStr)
```
jholdstock, line 111: Does this need to be decoded?

webapi/types.go
```diff
@@ -71,3 +71,16 @@ type ticketStatusResponse struct {
 	VoteChoices     map[string]string `json:"votechoices"`
 	Request         []byte            `json:"request"`
 }
+
+type SetAltSigRequest struct {
+	Timestamp     int64  `json:"timestamp" binding:"required"`
+	TicketHash    string `json:"tickethash" binding:"required"`
+	TicketHex     string `json:"tickethex" binding:"required"`
+	ParentHex     string `json:"parenthex" binding:"required"`
+	AltPubAddress string `json:"altpubaddress"`
```
jholdstock, line 80: Should this be `AltSigAddress`? And shouldn't it be required?

database/altsig.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// InsertAltSig will insert the provided ticket into the database. Returns an
+// error if either the ticket hash already exist.
```
jholdstock, line 39:
```suggestion
// error if the ticket hash already exists.
```

webapi/setaltsig.go
```diff
@@ -0,0 +1,137 @@
...
+	if currentAltSigAddr != "" {
+		msg := "alternate signature address already exists"
+		log.Errorf("%s: %s (ticketHash=%s)", funcName, msg, ticketHash)
+		sendErrorWithMsg(msg, errBadRequest, c)
+		return
+
+	}
+
+	// Disallow empty addresses.
+	if altPubAddr == "" {
```
jholdstock, line 68: This will be unnecessary if we add json annotations to make it required.

jholdstock (changes requested):

webapi/types.go
JoeGruffins: Ah, yes.

webapi/setaltsig.go
JoeGruffins: Apparently? It is a string in base64 in the header it seems. Is wrong here? Is working tho.

webapi/middleware.go
JoeGruffins: I think you are correct. Will try.

database/altsig.go
JoeGruffins: Ok.

webapi/setaltsig.go
JoeGruffins: That hadn't occurred to me as something necessary. We could always spoof that later, right? We have the keys. Does it make something more secure here? Will add though.

webapi/setaltsig.go
```diff
@@ -0,0 +1,137 @@
...
+		sendError(errInternalError, c)
+		return
+	}
+	history := &database.AltSigHistory{
+		Addr: sigAddress,
+		Req:  reqBytes,
+		Sig:  sig,
+	}
+
+	err = db.InsertAltSig(ticketHash, altPubAddr, history)
```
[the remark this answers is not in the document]
JoeGruffins: It made sense when multiple histories could be added, but you are right, they can be combined now. Will do so.

webapi/setaltsig.go
JoeGruffins: Or you mean just store as string? That's fine...

webapi/setaltsig.go
JoeGruffins: Ok.

webapi/setaltsig.go
JoeGruffins: Saving as string.

webapi/middleware.go
JoeGruffins: Yeah, now that we can only set once and not clear, it can only ever be the commitment address.

JoeGruffins:
I think I got everything commented upon. Tested buying a few more tickets and looks ok so far.


how do you like this updating style? It's my first time to try, and sorta my idea I guess. Is it confusing? It feels more organized for me, keeping changes together, but it may be just me. Perhaps quoting the commit hash in replies?

database/altsig.go
```diff
@@ -0,0 +1,126 @@
...
+			Addr: string(bkt.Get(altSigHistoryAddrK)),
+			Req:  bkt.Get(altSigHistoryReqK),
+			Sig:  bkt.Get(altSigHistorySigK),
+		}
+		return nil
+	})
+}
+
+// AltSigAddr retrieves a ticket's alternate signature address. Can be used to
+// check for the existence of an altsig by comparing to a null string.
```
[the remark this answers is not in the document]
JoeGruffins: 9aebd65

webapi/setaltsig.go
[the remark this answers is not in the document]
JoeGruffins: 9aebd65 and others

webapi/setaltsig.go
JoeGruffins: 9aebd65 and 0175871

database/altsig.go
JoeGruffins: 9aebd65

webapi/middleware.go
JoeGruffins: 9aebd65, 0175871, and c709561

webapi/setaltsig.go
JoeGruffins: 9aebd65

webapi/types.go
JoeGruffins: 0175871
Also changed in the decrediton pr so it works.

database/altsig.go
JoeGruffins: 9aebd65

webapi/setaltsig.go
JoeGruffins: 0175871

webapi/setaltsig.go
jholdstock: You're correct to say that a server can forge a response at will, at any time.

Hmm. Trying to imagine a scenario where it would make sense.

1. Client sends a request to set voting preferences (X). Server encounters an error, does not update database, provides an error response to client.
1. As a result, server votes with wrong preferences (Y).
1. Client makes a complaint.
1. Admin investigates. In their DB they see the request which set the X preferences, and they know that their server voted Y, but they have no way of knowing that the response was an error. Confusion and/or argument ensue.

Maybe this is a bit contrived? I think we originally justified storing the response by thinking that a complete audit trail would be more valuable than just half the trail, and the cost of storing the responses is relatively cheap.

jholdstock:
The update style seems good to me, its easy to follow. 

Which part of the nginx config is required exactly? Theres a lot there. I'm guessing something to do with ssl?

webapi/setaltsig.go
JoeGruffins: Yeah, it's probably good accounting. No problem saving it. Just wondering if there was something important I was missing.

JoeGruffins:
> Which part of the nginx config is required exactly? Theres a lot there. I'm guessing something to do with ssl?

Specifically the part under `if ($request_method = 'OPTIONS') {` I belive. It isn't optomized. There may be something extra I should be doing in decrediton. Will look into in more unless it's something you know about...

webapi/middleware.go
```diff
@@ -305,58 +305,92 @@ func vspAuth() gin.HandlerFunc {
...
+		// the request body with private key corresponding to sAddr.
+		// Verification failure results in the client getting notified
+		// of the cause. Returns whether verification succeeded.
+		validate := func(kt bool) {
+			t := &ticket
+			validateSig := func(addr string) bool {
+				// Validate request signature to ensure ticket ownership.
+				err = validateSignature(reqBytes, addr, c)
+				if err != nil {
+					log.Warnf("%s: Bad signature (clientIP=%s, ticketHash=%s): %v", funcName, c.ClientIP(), hash, err)
```
jholdstock, line 329: If I am reading correctly, a request which has an altSig in the database but has been signed with the commitment address will result in two log lines:

1. From failed altSig validation `Warn: Bad Signature`
1. From successful cmtAddr validation `Trace: Authed`

The warning seems unnecessary, nothing is wrong in this scenario. We only need a warning if both fail.

database/altsig.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"errors"
+	"fmt"
+
+	bolt "go.etcd.io/bbolt"
+)
+
+// ErrMaxAltSigs is returned when attempting to add a ticket's alternate
+// signature with one already in the database.
+var ErrMaxAltSigs = errors.New("max alt sigs")
```
jholdstock, line 16: This is no longer used

database/altsig.go
```diff
@@ -0,0 +1,137 @@
...
+	altSigAddrK = []byte("altsig")
+	reqSigAddrK = []byte("reqsigaddr")
+	reqK        = []byte("req")
+	reqSigK     = []byte("reqsig")
+	resK        = []byte("res")
+	resSigK     = []byte("ressig")
+)
+
+// AltSigData holds the information needed to prove that a client changed
+// their alternate signature address.
```
jholdstock, line 29:
```suggestion
// AltSigData holds the information needed to prove that a client added
// an alternate signature address.
```

docs/api.md
```diff
@@ -65,6 +65,40 @@ when a VSP is closed will result in an error.
 **Registering a ticket is a two step process. The VSP will not add a ticket to
 its voting wallets unless both of these calls have succeeded.**
 
+#### Step Zero (optional)
+
+Set an alternate signing address for a ticket. The ticket must be valid and will
+be transmitted to the network if not found. If set, the vsp will check that a
```
jholdstock, line 71: > If set, the vsp will check that a signature is good for this address and fallback to the commitment address if not.

I guess this is leftover from when alt sig could be updated, because it doesnt seem to make sense any more

docs/api.md
```diff
@@ -65,6 +65,40 @@ when a VSP is closed will result in an error.
...
+
+    Request:
+
+    ```json
+    {
+        "timestamp":1590509066,
+        "tickethash":"1b9f5dc3b4872c47f66b148b0633647458123d72a0f0623a90890cc51a668737",
+        "tickethex":"0100000001a8...bfa6e4bf9c5ec1",
+        "parenthex":"0100000022a7...580771a3064710",
+        "altpubaddress":"Tsfkn6k9AoYgVZRV6ZzcgmuVSgCdJQt9JY2"
```
jholdstock, line 88: altsigaddress*

docs/api.md
```diff
@@ -65,6 +65,40 @@ when a VSP is closed will result in an error.
 **Registering a ticket is a two step process. The VSP will not add a ticket to
 its voting wallets unless both of these calls have succeeded.**
 
+#### Step Zero (optional)
+
+Set an alternate signing address for a ticket. The ticket must be valid and will
+be transmitted to the network if not found. If set, the vsp will check that a
+signature is good for this address and fallback to the commitment address if not.
+The address must be valid for the network and a pay to secp256k1 ecdsa pubkey
```
jholdstock, line 73:
```suggestion
The address must be valid for the network and a pay to secp256k1 ECDSA pubkey
```

jholdstock (changes requested):
Fully reviewed all apart from test code.

JoeGruffins:
Thanks for the great review! Will squash outstanding *squash* commits before addressing.

JoeGruffins:
hmm, that was a botched rebase

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Fail fast if the pubkey doesn't decode properly.
+	addr, err := stdaddr.DecodeAddressV0(altSigAddr, cfg.NetParams)
+	if err != nil {
+		log.Tracef("%s: Alt sig address cannot be decoded (clientIP=%s): %v", funcName, c.ClientIP(), err)
+		sendErrorWithMsg(err.Error(), errBadRequest, c)
+		return
+	}
+	if _, ok := addr.(*stdaddr.AddressPubKeyHashEcdsaSecp256k1V0); !ok {
+		log.Tracef("%s: Alt sig address is unexpected type (clientIP=%s type=%T): %v", funcName, c.ClientIP(), addr, err)
```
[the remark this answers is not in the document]
JoeGruffins: But `*stdaddr.AddressScriptHashV0` is the wrong type and is correct to log? For exaple: https://play.golang.org/p/2rFLzwSAVM2

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+
+	sig := ed25519.Sign(signPrivKey, resBytes)
+	serverSigStr := base64.StdEncoding.EncodeToString(sig)
+	c.Writer.Header().Set("VSP-Server-Signature", serverSigStr)
+
+	c.AbortWithStatusJSON(http.StatusOK, res)
+
+	data := &database.AltSigData{
+		AltSigAddr: altSigAddr,
+		ReqSigAddr: commitmentAddr,
```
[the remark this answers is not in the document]
JoeGruffins: The commitment address was used to sign Req which adds the alternate signature address. It must be known in order to prove that the user added an alternate signature.

> all we really need for vspd to function is a mapping between ticketHash and altSig

This is true as long as we can fish out the commitment address later. Since I believe we always can, we can remove this. It should be in the ticket hex in the request.

docs/api.md
JoeGruffins: I think this is still true. It will check if set and fallback to the commitment address.

database/altsig.go
JoeGruffins: Removed in d5595c9

database/altsig.go
JoeGruffins: Changed in d5595c9

database/upgrade_v4.go
```diff
@@ -0,0 +1,41 @@
...
+			return fmt.Errorf("failed to update db version: %w", err)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	log.Info("Upgrade v4 completed")
```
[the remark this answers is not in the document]
JoeGruffins: Changed in d5595c9

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+	commitmentAddr := c.MustGet("CommitmentAddress").(string)
+
+	if cfg.VspClosed {
+		sendError(errVspClosed, c)
+		return
+	}
+
+	var request SetAltSigRequest
+	if err := binding.JSON.BindBody(reqBytes, &request); err != nil {
+		log.Tracef("%s: Bad request (clientIP=%s): %v", funcName, c.ClientIP(), err)
```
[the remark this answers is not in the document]
JoeGruffins: Changed in 003f5d0

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+
+	currentData, err := db.AltSigData(ticketHash)
+	if err != nil {
+		log.Errorf("%s: db.AltSigData (ticketHash=%s): %v", funcName, ticketHash, err)
+		sendError(errInternalError, c)
+		return
+	}
+	if currentData != nil {
+		msg := "alternate signature data already exists"
+		log.Errorf("%s: %s (ticketHash=%s)", funcName, msg, ticketHash)
```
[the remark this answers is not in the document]
JoeGruffins: Changed in 003f5d0

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+		sendErrorWithMsg(msg, errBadRequest, c)
+		return
+
+	}
+
+	// Fail fast if the pubkey doesn't decode properly.
+	addr, err := stdaddr.DecodeAddressV0(altSigAddr, cfg.NetParams)
+	if err != nil {
+		log.Tracef("%s: Alt sig address cannot be decoded (clientIP=%s): %v", funcName, c.ClientIP(), err)
```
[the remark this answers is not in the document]
JoeGruffins: Changed in 003f5d0

webapi/setaltsig.go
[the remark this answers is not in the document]
JoeGruffins: Other problems fixed in 003f5d0

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Ensure this ticket is eligible to vote at some point in the future.
+	canVote, err := dcrdClient.CanTicketVote(rawTicket, ticketHash, cfg.NetParams)
+	if err != nil {
+		log.Errorf("%s: dcrd.CanTicketVote error (ticketHash=%s): %v", funcName, ticketHash, err)
+		sendError(errInternalError, c)
+		return
+	}
+	if !canVote {
+		log.Tracef("%s: unvotable ticket (clientIP=%s, ticketHash=%s)",
```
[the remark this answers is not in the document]
JoeGruffins: Fixed in 003f5d0

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+		return
+	}
+
+	sigStr := c.GetHeader("VSP-Client-Signature")
+
+	res := SetAltSigResponse{
+		Timestamp: time.Now().Unix(),
+		Request:   reqBytes,
+	}
+	resBytes, err := json.Marshal(res)
```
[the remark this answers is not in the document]
JoeGruffins: Done in 003f5d0

webapi/setaltsig.go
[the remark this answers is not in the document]
JoeGruffins: Removed in 0441516 and ce1b625

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+		ReqSigAddr: commitmentAddr,
+		Req:        reqBytes,
+		ReqSig:     sigStr,
+		Res:        resBytes,
+		ResSig:     serverSigStr,
+	}
+
+	err = db.InsertAltSig(ticketHash, data)
+	if err != nil {
+		log.Tracef("%s: db.InsertAltSig error, failed to set alt history (ticketHash=%s): %v",
```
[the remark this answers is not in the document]
JoeGruffins: Fixed in 003f5d0

webapi/setaltsig.go
```diff
@@ -0,0 +1,142 @@
...
+			funcName, ticketHash, err)
+		sendError(errInternalError, c)
+		return
+	}
+
+	log.Tracef("%s: New alt sig pubkey set for ticket: (ticketHash=%s)", funcName, ticketHash)
```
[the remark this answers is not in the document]
JoeGruffins: Fixed in 003f5d0

webapi/middleware.go
```diff
@@ -305,58 +305,92 @@ func vspAuth() gin.HandlerFunc {
 			return
 		}
```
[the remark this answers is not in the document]
JoeGruffins: Adjusted slightly in 95d7a0a 
Needs more?

webapi/middleware.go
```diff
@@ -305,58 +305,92 @@ func vspAuth() gin.HandlerFunc {
...
-				log.Errorf("%s: dcrd.GetRawTransaction for ticket failed (ticketHash=%s): %v", funcName, hash, err)
-				sendError(errInternalError, c)
-				return
+		// validate validates the signature in the request as signing
+		// the request body with private key corresponding to sAddr.
+		// Verification failure results in the client getting notified
+		// of the cause. Returns whether verification succeeded.
+		validate := func(kt bool) {
+			t := &ticket
+			validateSig := func(addr string) bool {
```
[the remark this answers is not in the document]
JoeGruffins: Attempted to make more straightforward in 95d7a0a but if you have anything specific in mind please let me know.

webapi/middleware.go
JoeGruffins: Made so it doesn't fuss about the first error. Should I use more logic to maybe warn that both failed in case they do? It will only warn for the second one now.  95d7a0a

docs/api.md
JoeGruffins: Fixed in 0cc4a5b

docs/api.md
JoeGruffins: Fixed in 0cc4a5b

JoeGruffins:
Log levels were all over the place, sorry. Wasn't paying enough attention to those.

webapi/setaltsig_test.go
```diff
@@ -0,0 +1,236 @@
...
+// characters from the provided charset.
+func randString(length int, charset string) string {
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(b)
+}
+
+func tAltSigData() *database.AltSigData {
```
jholdstock, line 101: `exampleAltSigData`

webapi/setaltsig_test.go
```diff
@@ -0,0 +1,236 @@
...
+	// Set up some global params.
+	cfg.NetParams = chaincfg.MainNetParams()
+	_, signPrivKey, _ = ed25519.GenerateKey(seededRand)
+
+	// Create a database to use.
+	// Ensure we are starting with a clean environment.
+	os.Remove(testDb)
+	os.Remove(backupDb)
+
+	// Create a new blank database for each sub-test.
```
jholdstock, line 65: This isn't happening. One database is created, `m.Run()` runs all of the tests, and then the database is torn down once at the end.

webapi/setaltsig_test.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+		b, err := json.Marshal(req)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		handle := func(c *gin.Context) {
+			c.Set("DcrdClient", tNode)
+			c.Set("RequestBytes", b[test.deformReq:])
+			c.Set("CommitmentAddress", randString(35, addrCharset))
```
jholdstock, line 216: `"CommitmentAddress"` no longer required

webapi/setaltsig_test.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		reqSig := randString(504, hexCharset)
+		c.Request.Header.Set("VSP-Client-Signature", reqSig)
+
+		r.ServeHTTP(w, c.Request)
+
+		if test.wantCode != w.Code {
+			t.Errorf("%q: expected status %d, got %d", test.name, test.wantCode, w.Code)
+		}
```
jholdstock, line 234: Maybe an assertion at the end to check if the altsig has (or has not) been added to the database?

jholdstock (changes requested):
Last few minor things and I think we are good to go

database/altsig_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"testing"
+)
+
+const (
+	addrCharset = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+	hexCharset  = "1234567890abcdef"
+	sigCharset  = "0123456789ABCDEFGHJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz+/="
+)
+
+func randBytes(n int) []byte {
```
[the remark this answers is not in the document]
JoeGruffins: I think `randBytes` is only used in this file afaict. this test does use randString though, which is in the ticket_test file. Can move both to database_test tho.

database/altsig_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+package database
+
+import (
+	"reflect"
+	"testing"
+)
+
+const (
+	addrCharset = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
```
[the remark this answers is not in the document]
JoeGruffins: Grouped together in 32e2147

database/altsig_test.go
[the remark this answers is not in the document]
JoeGruffins: Grouped together in 32e2147

database/altsig_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+func randBytes(n int) []byte {
+	slice := make([]byte, n)
+	if _, err := seededRand.Read(slice); err != nil {
+		panic(err)
+	}
+	return slice
+}
+
+func tAltSigData() *AltSigData {
```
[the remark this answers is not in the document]
JoeGruffins: Renamed in 32e2147

database/altsig_test.go
```diff
@@ -0,0 +1,137 @@
...
+		t.Fatal("expected no data")
+	}
+
+	// Insert an altsig.
+	data := tAltSigData()
+	if err := db.InsertAltSig(ticketHash, data); err != nil {
+		t.Fatalf("unexpected error storing altsig in database: %v", err)
+	}
+
+	h, err = db.AltSigData(ticketHash)
```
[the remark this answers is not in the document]
JoeGruffins: Indeed. Done in 32e2147

webapi/setaltsig.go
```diff
@@ -0,0 +1,124 @@
...
+		AltSigAddr: altSigAddr,
+		Req:        reqBytes,
+		ReqSig:     sigStr,
+		Res:        []byte(res),
+		ResSig:     resSig,
+	}
+
+	err = db.InsertAltSig(ticketHash, data)
+	if err != nil {
+		log.Errorf("%s: db.InsertAltSig error, failed to set alt history (ticketHash=%s): %v",
```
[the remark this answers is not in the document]
JoeGruffins: Fixed in 32e2147

webapi/setaltsig_test.go
JoeGruffins: Fixed in 32e2147

webapi/setaltsig_test.go
JoeGruffins: Changed wording in 3727325

webapi/setaltsig_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+func (n *testNode) GetRawTransaction(txHash string) (*dcrdtypes.TxRawResult, error) {
+	return nil, n.getRawTransactionErr
+}
+
+func TestSetAltSig(t *testing.T) {
+	const testAddr = "DsVoDXNQqyF3V83PJJ5zMdnB4pQuJHBAh15"
+	tests := []struct {
+		name                 string
+		vspClosed, clearHash bool
```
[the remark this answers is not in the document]
JoeGruffins: Removed in 3727325

webapi/setaltsig_test.go
JoeGruffins: Removed in 3727325

webapi/setaltsig_test.go
JoeGruffins: Assertion added in 3727325

database/altsig_test.go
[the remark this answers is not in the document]
jholdstock: Ah yep, youre right, its only used in that file. To be honest, I care more about removing the duplication than about exactly where the function lives, so moving it into database_test is fine.

jholdstock (approved):

JoeGruffins:
Will squash squashables.

webapi/setaltsig.go
```diff
@@ -0,0 +1,124 @@
...
+		log.Warnf("%s: unvotable ticket (clientIP=%s, ticketHash=%s)",
+			funcName, c.ClientIP(), ticketHash)
+		sendError(errTicketCannotVote, c)
+		return
+	}
+
+	sigStr := c.GetHeader("VSP-Client-Signature")
+
+	// Send success response to client.
+	res, resSig := sendJSONResponse(SetAltSigResponse{
```
xaur, line 103: This could be a silly question (if this line means "queue" and not "send") but I'll take the beating for a chance to spot a problem.

Could there be an issue with sending success _before_ saving data to the db?

jholdstock: This does return a response instantly, its not queued. This is a good catch actually, I overlooked it because we do it in other places, but in those cases we are only storing an audit trail. This case is actually storing some important data, and if this db call fails we should return an error to the client. I will fix this.

PR #301, opened 2021-10-27
Record expired/missed tickets as revoked.

jholdstock:
This is an unfortunate workaround which is necessary because an instance of dcrwallet will only recognise a ticket as revoked if it has done the revoke itself. If another wallet broadcasts the revoke then the ticket will forever be reported as missed/expired. This causes problems for vspd when a wallet outside of the vspd deployment revokes a ticket (eg. a users ticketbuyer wallet).

I'm happy to include this slightly dirty workaround because this should no longer be an issue when the auto-revoke work in DCP-0009 lands.

### DIFF
--- a/background/background.go
+++ b/background/background.go
@@ -253,28 +253,28 @@ func blockConnected() {
 	// successful wallet will have the most up-to-date ticket status, the others
 	// will be outdated.
 	for _, walletClient := range walletClients {
-		dbTickets, err := db.GetVotableTickets()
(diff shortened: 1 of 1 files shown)


jrick:
I'm fuzzy on these details as well.  It sounds like some sort of tx filtering issue (in RPC mode), where wallets are not informed of the revocation?

I don't see anything wrong with this workaround, but the wallet bug is a little surprising to me.

PR #307, opened 2021-11-14
Clarify "setaltsig" terminology.

jholdstock:
Stardardize `alt sig`/`alt signature`/`alt signing address` terminology to `alternate signing address`.

As discussed with @xaur and @JoeGruffins https://github.com/decred/vspd/issues/279#issuecomment-962380648


database/altsignaddr.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"errors"
+	"fmt"
+
+	bolt "go.etcd.io/bbolt"
+)
+
+// The keys used to store alternate signing addresses in the database.
+var (
+	altSignAddrK = []byte("altsig")
```
xaur, line 16: How about the key too?

database/database.go
```diff
@@ -50,8 +50,8 @@ var (
 	privateKeyK = []byte("privatekey")
 	// lastaddressindex is the index of the last address used for fees.
 	lastAddressIndexK = []byte("lastaddressindex")
-	// altSigBktK stores alternate signatures.
-	altSigBktK = []byte("altsigbkt")
+	// altSignAddrBktK stores alternate signing addresses.
+	altSignAddrBktK = []byte("altsigbkt")
```
xaur, line 54: How about the bucket name too?

xaur (commented):
More characters but way less confusing now!

database/altsignaddr.go
jholdstock: Don't want to change the keys because it will break compatibility any existing databases, and tbh the values of the keys don't really matter at all.

The proper way to change a db key would be to write code which performs a database migration, and in this case I just don't think its worth it.

database/database.go
jholdstock: See previous

database/altsignaddr.go
xaur: That was my guess too, asked just in case. Thanks!

xaur:
We don't have any downstream clients to notify of this change, like WIP patches, right?

jholdstock:
@JoeGruffins is aware of the dcrwallet/decrediton changes for this.

dnldd (approved):

PR #311, opened 2021-12-28
Implement Admin Password hashing using sha256

ukane-philemon:
Per [SEI CERT C Coding Standard](https://wiki.sei.cmu.edu/confluence/display/c/MSC18-C.+Be+careful+while+handling+sensitive+data,+such+as+passwords,+in+program+code), it is best practice not to store plain text passwords in memory or on disk. This was achieved by storing the sha256 hash of the admin pass, removing the provided password bytes from memory, and constant-time comparing the hashes of incoming passwords with the stored hash of the admin pass. For input on startup, the secure terminal is used to input the admin pass every start-up. Closes decred#281


degeri:
Just wondering that this would impact auto recovery systems. If a system crashes it will require admin to manually start the service for a secure setup (Imagine a long Christmas break and the API being down for days). Why not just store the hash to a config file in disk after the first prompt ? This would provide almost the same level of security without the need to prompt the user for password on each startup (Hence boot scripts can be automated)

jholdstock:
Hi @ukane-philemon. Definitely appreciate the PR but I have to agree with @degeri here. vspd is designed to be run as a system service which can be started/restarted non-interactively. 

I notice in [this PR](https://github.com/itswisdomagain/vspd/pull/8) you used the database to store the password hash, and that seems like a good idea. Why did you change it? A better solution could be to just store it in a new file in the datadir (eg. `~/.vspd/password.hash`). That would make forgotten password recovery very easy - just delete the file and set a new password through the terminal. It would also allow admins to easily install the file using third party tools like ansible.

Also, in that PR you used bcrypt, and this one is using SHA-256. Why the change?

ukane-philemon:
@degeri @jholdstock thank you for looking into this PR, it's my first time contributing here so I was concerned about what the team would accept hence my switch to sha256 as it is used in dcrdex, except dcrstakepool which used bcrypt.
My initial PR used bcrypt and saved admin password to db. Operators can easily change the password without the need to delete a file in their datadir, by providing it via terminal or config.
If it's ok, I can close this PR and submit the other?

ukane-philemon:
@jholdstock @degeri I know projects have their preferences when it comes to hashing.
SHA-256 or Bcrypt?
Bcrypt was especially made for passwords but SHA-256 use less code to do what it does.

ukane-philemon:
@jholdstock I resolved to using bcrypt in my recent [commit](https://github.com/decred/vspd/pull/311/commits/eed3ef781a8aeae058a50f4b3d7166d75e9f7fa3). The [bcrypt password hashing](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html) mechanism is the second choice after Argon2id, for password hashing/storage to achieve [FIPS-140 compliance](https://en.wikipedia.org/wiki/FIPS_140-2).

degeri:
Yeah should not be too big of a difference. Lets go with SHA since its less dependencies

jholdstock:
While I agree with the premise of this change, and the code looks acceptable, I'd like to hold off merging this PR for a while because release 1.1.0 is imminent (to coincide with dcrd/dcrwallet 1.7).

Changing vspd from a non-interactive to an interactive process is a very significant change to make so close to a release.

ukane-philemon:
@jholdstock @degeri thanks for your feedback and review of this piece. I'm happy it's slated for v1.1.

PR #313, opened 2022-01-07
Complete all DB operations before returning responses to client.

### LINKED ISSUE #304: Complete all DB operations before returning responses
This could be a silly question (if this line means "queue" and not "send") but I'll take the beating for a chance to spot a problem.

Could there be an issue with sending success _before_ saving data to the db?

_Originally posted by @xaur in https://github.com/decred/vspd/pull/287#discussion_r744060677_

ukane-philemon:
This commit ensures all DB operations are successfully completed before sending success response to client. If DB operations are unsuccessful, an error message is sent. Closes #304


jholdstock:
Closing per conversation in #304

PR #314, opened 2022-01-07
Add alternate signing address to /ticketstatus response

### LINKED ISSUE #306: Return alternate signing address in /ticketstatus response
https://github.com/decred/vspd/blob/master/docs/api.md#ticket-status

@JoeGruffins do you think this is important? Could maybe be useful for recovering from error scenarios, or restoring a wallet from seed?

ukane-philemon:
This PR adds alternate signing address to /ticketstatus response.
Closes #306

### DIFF
--- a/webapi/ticketstatus.go
+++ b/webapi/ticketstatus.go
@@ -34,12 +34,26 @@ func ticketStatus(c *gin.Context) {
 		return
 	}
 
+	// Get altSignAddress from database
+	altSignAddrData, err := db.AltSignAddrData(ticket.Hash)
+	if err != nil {
+		log.Errorf("%s: db.AltSignAddrData error (ticketHash=%s): %v", funcName, ticket.Hash, err)
+		sendError(errInternalError, c)
+		return
+	}
+
+	altSignAddr := ""
+	if altSignAddrData != nil {
+		altSignAddr = altSignAddrData.AltSignAddr
+	}
+
 	sendJSONResponse(ticketStatusResponse{
 		Timestamp:       time.Now().Unix(),
 		Request:         reqBytes,
 		TicketConfirmed: ticket.Confirmed,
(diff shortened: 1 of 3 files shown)


webapi/types.go
```diff
@@ -68,6 +68,7 @@ type ticketStatusResponse struct {
 	TicketConfirmed bool              `json:"ticketconfirmed"`
 	FeeTxStatus     string            `json:"feetxstatus"`
 	FeeTxHash       string            `json:"feetxhash"`
+	AltSignAddress  string            `json:"altsignaddress"`
```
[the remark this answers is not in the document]
jholdstock: I actually prefer `altsignaddress` here because it is consistent with the rest of the code. We don't use the term `altsigningaddress` anywhere else.

webapi/ticketstatus.go
```diff
@@ -34,12 +34,24 @@ func ticketStatus(c *gin.Context) {
 		return
 	}
 
+	// Get altSignAddress from database
+	altSignAddrData, err := db.AltSignAddrData(ticket.Hash)
+	if err != nil {
+		log.Errorf("db.AltSignAddrData error (ticketHash=%s): %v", ticket.Hash, err)
```
jholdstock, line 40: Log line should be prefixed with `funcName`.

jholdstock, line 40: I think we probably want to return an error to the client if a database error happens here.

jholdstock (changes requested):

webapi/ticketstatus.go
jholdstock: An example where we already call `db.AltSignAddrData`

https://github.com/decred/vspd/blob/0fd96388cecd6506174ba9bc33d6e77d42a5fc90/webapi/setaltsignaddr.go#L57-L62

jholdstock (approved):

PR #317, opened 2022-01-17
Use DCP-0010 status to calculate fees

jholdstock:


rpc/dcrd.go
```diff
@@ -140,6 +141,26 @@ func (c *DcrdRPC) SendRawTransaction(txHex string) error {
...
+
+	agenda, ok := info.Deployments[chaincfg.VoteIDChangeSubsidySplit]
+	if !ok {
+		return false, fmt.Errorf("getblockchaininfo did not return agenda %q",
+			chaincfg.VoteIDChangeSubsidySplit)
+	}
+
+	activeState := blockchain.ThresholdStateTuple{State: blockchain.ThresholdActive}
+
+	return agenda.Status == activeState.String(), nil
```
jholdstock, line 161: Rather than comparing status strings, perhaps a better way to do this would be `agenda.Since < bestBlock.Height`?

rpc/dcrd.go
```diff
@@ -140,6 +141,26 @@ func (c *DcrdRPC) SendRawTransaction(txHex string) error {
 	return nil
 }
 
+// DCP0010Active uses getblockchaininfo RPC to determine if the DCP-0010 agenda
+// has activated on the current network.
+func (c *DcrdRPC) DCP0010Active() (bool, error) {
```
dnldd, line 146: Wouldn't `DCP0010Active` be a better name here?

webapi/getfeeaddress.go
```diff
@@ -19,6 +19,8 @@ import (
 // addrMtx protects getNewFeeAddress.
 var addrMtx sync.Mutex
 
+const defaultMinRelayTxFee = dcrutil.Amount(1e4)
```
dnldd, line 22: Since the constant is only being used in `getCurrentFee` the scope could be reduced to just that function?

dnldd (changes requested):

rpc/dcrd.go
dnldd: Isn't the status comparison a lot more clearer?

rpc/dcrd.go
jholdstock: Umm... that is the name?

rpc/dcrd.go
jholdstock: It is easier to read, but it involves setting up the `TresholdStateTuple` each time which is a bit verbose and less performant. The tuple cannot be set as a constant so needs to be created each time the func is called.

davecgh: I don't think so as you'd still need to check which status is being reported.  The `.Since` field is set for `lockedin` as well.

In fact, this really needs to check two conditions since votes apply the previous block and thus if you call `getblockchaininfo` at the tip height at the block just before the agenda activates, it will not be active yet and thus this would return false even though the votes will need to have the new subsidy.

I would suggest something like the following (dry coded, so maybe there is a typo, but logically correct):

```Go
var (
	activeStatus = blockchain.ThresholdStateTuple{State: blockchain.ThresholdActive}.String()
	lockedInStatus = blockchain.ThresholdStateTuple{State: blockchain.ThresholdLockedIn}.String()
)

func (c *DcrdRPC) DCP0010Active() (bool, error) {
	net := params.Net
	rcai := int32(params.RuleChangeActivationInterval)

	if net == wire.SimNet {
		return true, nil
	}
	if net != wire.MainNet && net != wire.TestNet3 {
		return false, nil
	}

	// Do the RPC....

	switch {
	case agenda.Status ==lockedInStatus && height == int32(d.Since)+rcai-1:
		return true, nil
	case agenda.Status == activeStatus:
		return true, nil
	}
	return false, nil
```

davecgh: I should also note you can technically skip the extra developer network checks if vspd is never to be used on those networks, but might as well be accurate.

jholdstock: Tuple could be set as a package global I suppose, but I wanted to avoid those because vspd has too many already.

davecgh: It's going to be inlined anyway since the Stringers just return constant strings.

rpc/dcrd.go
dnldd: Whoops, meant `IsDCP0010Active`.

rpc/dcrd.go
jholdstock: I've moved the status string to a package global so it doesn't need to be reconstructed on each function call.

---

I appreciate the nuance you are explaining here @davecgh, however in the context of vspd I don't think its worth the extra code/complexity to get this *exactly* right. vspd is not responsible for creating any vote transactions or dealing with any stakebases.

The value of `IsDCP0010Active` will only be used for calculating fee payments, and the fee payments are only an estimated value anyway. The code projects into the future to estimate when a ticket is likely to vote, and estimates what the vote reward at that time will be. The fee is then a percentage of that estimated value.

I briefly discussed this with @jrick - there is going to be an awkward transition period where users are purchasing tickets before the agenda activates, but the ticket votes after activation. In these cases, the user will have paid a fee based on the old vote reward, but they will receive the new vote reward. Good for the user, but the server admin will receive a slightly smaller fee than might be expected.

If this transition period is extended by one block makes a negligible difference.

dnldd (approved):

rpc/dcrd.go
davecgh: Seem fine as long as the server doesn't refuse to vote because it notices the fee is too small after the fact.  I'm not familiar with the exact mechanics of how vspd handles that aspect.  It probably just checks it once at the time of payment, and if that is the case, I agree it won't make a difference outside of what you identified in your assessment.

davecgh (approved):

rpc/dcrd.go
jholdstock: vspd doesn't do any voting, that's all on dcrwallet.

You're correct about it being a one time check - as long as the clients pay the requested fee, the deal is done. Their tickets are added to the voting wallets and remain there until spent.

As always, thanks for the details review and discussion.

PR #318, opened 2022-01-24
Using error.Is and error.As

ukane-philemon:
This diff allows using error.Is and error.As within vsp, instead of error string comparison. This is useful in situations where an error message sent to client is based on the kind of apiError returned from a helper function within vsp.

### DIFF
--- a/webapi/errors.go
+++ b/webapi/errors.go
@@ -73,8 +73,8 @@ func (e apiError) httpStatus() int {
 	}
 }
 
-// defaultMessage returns a descriptive error string for a given error code.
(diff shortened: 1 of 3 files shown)


jholdstock:
Great idea, thanks for implementing. Would love to see some tests for this to ensure that `.As` and `.Is` work as expected. 

A good example of this can be found in dcrd: https://github.com/decred/dcrd/blob/master/gcs/error_test.go

ukane-philemon:
@jholdstock thanks for your feedback. I have written some test case for the new implementation.

PR #324, opened 2022-02-09
votingwalletsonline and totalvotingwallets stats

### LINKED ISSUE #272: Health API
I suggest to discuss the addition of health metrics to existing endpoint like [`/vspinfo`](https://ultravsp.uk/api/v3/vspinfo) or create a new endpoint.

Two metrics to start with are `votingwalletsonline` and `votingwalletstotal`. I see the following uses for these fields:

- when online is lower than the total, wallets could alert the user that there is an issue with the VSP
- wallets could also present `votingwalletstotal` as one data point reflecting the reliability of a given VSP. iirc some dcrstakepool instances settled on the minimum required 3 wallets, while some added more have 5 or even 8 voting wallets
- both motivations above are also valid for automated health monitor bots that could build reports or dashboards of the VSP ecosystem (see https://github.com/decredcommunity/issues/issues/148 for more on this)

This idea is inspired by one API I used that returned the count of shards online and shards total. These allowed me to alert the user that the data returned may be incomplete.

Extracted from #232.

ukane-philemon:
This adds new stats `votingwalletsonline` and `totalvotingwallets` to `/vspinfo` endpoint. closes #272

### DIFF
--- a/webapi/cache.go
+++ b/webapi/cache.go
@@ -20,15 +20,17 @@ import (
 // apiCache is used to cache values which are commonly used by the API, so
 // repeated web requests don't repeatedly trigger DB or RPC calls.
 type apiCache struct {
-	UpdateTime        string
-	PubKey            string
-	DatabaseSize      string
-	Voting            int64
-	Voted             int64
-	Revoked           int64
-	BlockHeight       uint32
-	NetworkProportion float32
-	RevokedProportion float32
+	UpdateTime          string
+	PubKey              string
+	DatabaseSize        string
+	Voting              int64
+	Voted               int64
+	Revoked             int64
+	VotingWalletsOnline int64
+	TotalVotingWallets  int64
+	BlockHeight         uint32
+	NetworkProportion   float32
+	RevokedProportion   float32
 }
 
 var cacheMtx sync.RWMutex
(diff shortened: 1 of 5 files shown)


webapi/cache.go
```diff
@@ -83,13 +85,23 @@ func updateCache(ctx context.Context, db *database.VspDatabase,
 		return errors.New("dcr node reports a network ticket pool size of zero")
 	}
 
+	clients, failedConnections := wallets.Clients(ctx, cfg.NetParams)
+	if len(clients) == 0 {
+		return errors.New("Could not connect to any wallets")
```
jholdstock, line 90: We should keep the error log if `len(failedConnections) > 0`, but I see no reason to return an error if `len(clients) == 0`. We aren't actually using the wallet clients for anything at this time, just checking if they are online, and we still want to update the cached information.

jholdstock (changes requested):

jholdstock:
This is good to merge once the error return is removed.

webapi/cache.go
ukane-philemon: Thanks for the review, fixed this in the latest fp.

jholdstock (approved):

PR #325, opened 2022-02-09
Support link to legacy vsp

### LINKED ISSUE #224: Add link to legacy dcrstakepool instance
The tabs at [stakey.net](https://stakey.net/) to both old and new stats look useful. Migration to vspd will take a while (I project ~1 year) and it would help to link things together until it is over.

I'm not sure where to place the link. A tab would be nice. If not, at least add a placeholder in the part of the home page that is customizable by the operator.

The idea is to help VSP operators to establish hyperlinks between dcrstakepool and vspd sites with minimum hacking of their home pages.

ukane-philemon:
This makes it possible for VSP operators establish hyperlinks with their dcrstakepool site. closes #224 
PS: @jholdstock placing a single tab would look lonely on the navigation, what do you think?
If link to legacy vsp is provided:
legacylink


jholdstock:
Thanks @ukane-philemon but I don't think we will merge this. It might have made sense in the initial release of vspd, but now only  0.3% of tickets are held by legacy VSPs, so I dont think it makes sense to keep linking to them in new code.

PR #327, opened 2022-03-15
Remove global vars from webapi package.

jholdstock:
Just code refactoring here, no functional changes. Removing global vars from the webapi package and instead instantiating a `Server` to hold those variables.


dajohi (approved):

jholdstock:
Contributes to #339

PR #328, opened 2022-03-16
Use GitHub mermaid.js rendering.

jholdstock:
GitHub is now capable of rendering mermaid.js diagrams, so we don't need to maintain a separate source and rendered .png file. The mermaid source can be embedded directly into the documentation itself.


chappjc (approved):
Very cool.  Renders fine for me, although with a new JavaScript domain, viewscreen.githubusercontent.com.

dajohi (approved):

PR #322, opened 2022-02-01
Add manual ticket search feature.

ukane-philemon:
This commit adds manual ticket search feature to the vspd frontend. Tickets owners can search tickets with a base64 encoded string containing the ticket hash and signature. if ownership validation is successful, ticket owners can view full information about their ticket. This aids confirming if ticket registration is successful without depending on decrediton or dcrwallet verifying via an api. closes #227 
Ticket page:
newlook2
Homepage before: 
homepagebefore
Modified Homepage:
newhomepage

PR #329, opened 2022-03-16
add new helper functions

ukane-philemon:
This diiff adds new helper funcs `validateTicketHash()` and `getCommitmentAddress()`. Also it modifies helper func `validateSignature()`. Originally part of #322

### DIFF
--- a/webapi/helpers.go
+++ b/webapi/helpers.go
@@ -15,6 +15,7 @@ import (
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/decred/dcrd/dcrutil/v4"
 	"github.com/decred/dcrd/wire"
+	"github.com/decred/vspd/rpc"
 	"github.com/gin-gonic/gin"
 )
 
@@ -105,16 +106,22 @@ func validPolicyOption(policy string) error {
 	}
 }
 
-func validateSignature(reqBytes []byte, commitmentAddress string, c *gin.Context) error {
-	// Ensure a signature is provided.
-	signature := c.GetHeader("VSP-Client-Signature")
-	if signature == "" {
-		return errors.New("no VSP-Client-Signature header")
-	}
+func validateSignature(hash, commitmentAddress, signature, message string, c *gin.Context) error {
+	firstErr := dcrutil.VerifyMessage(commitmentAddress, signature, message, cfg.NetParams)
+	if firstErr != nil {
+		// Don't return an error straight away if sig validation fails -
+		// first check if we have an alternate sign address for this ticket.
+		altSigData, err := db.AltSignAddrData(hash)
+		if err != nil {
+			return fmt.Errorf("db.AltSignAddrData failed (ticketHash=%s): %v", hash, err)
+		}
 
-	err := dcrutil.VerifyMessage(commitmentAddress, signature, string(reqBytes), cfg.NetParams)
-	if err != nil {
-		return err
+		// If we have no alternate sign address, or if validating with the
+		// alt sign addr fails, return an error to the client.
+		if altSigData == nil || dcrutil.VerifyMessage(altSigData.AltSignAddr, signature, message, cfg.NetParams) != nil {
+			return fmt.Errorf("Bad signature (clientIP=%s, ticketHash=%s)", c.ClientIP(), hash)
+		}
+		return firstErr
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


webapi/helpers.go
```diff
@@ -141,3 +148,59 @@ func isValidTicket(tx *wire.MsgTx) error {
 
 	return nil
 }
+
+// validateTicketHash ensures the provided ticket hash is a valid ticket hash.
+// A ticket hash should be 64 chars (MaxHashStringSize) and should parse into
+// a chainhash.Hash without error.
+func validateTicketHash(c *gin.Context, hash string) (bool, error) {
```
jholdstock, line 155: This func doesn't need to return a bool. The bool doesn't give any extra information versus just checking `if err != nil` (it actually gives less, because there is no descriptive error message).

See `isValidTicket` for a similar pattern where only an error is returned.

webapi/helpers.go
```diff
@@ -141,3 +148,59 @@ func isValidTicket(tx *wire.MsgTx) error {
...
+	if len(hash) != chainhash.MaxHashStringSize {
+		return false, fmt.Errorf("Incorrect hash length (clientIP=%s): got %d, expected %d", c.ClientIP(), len(hash), chainhash.MaxHashStringSize)
```
jholdstock, line 157: Error strings should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

This should have been caught by the linter in CI, I am investigating why that didnt happen.

webapi/helpers.go
```diff
@@ -141,3 +148,59 @@ func isValidTicket(tx *wire.MsgTx) error {
...
+	}
+
+	return true, nil
+}
+
+// getCommitmentAddress gets the commitment address of the provided ticket hash
+// from the chain.
+func getCommitmentAddress(c *gin.Context, hash string) (string, bool, error) {
+	var commitmentAddress string
+	dcrdClient := c.MustGet(dcrdKey).(*rpc.DcrdRPC)
```
jholdstock, line 173: I think `dcrdClient` should be passed into this function as a param.

webapi/helpers.go
```diff
@@ -141,3 +148,59 @@ func isValidTicket(tx *wire.MsgTx) error {
...
+// A ticket hash should be 64 chars (MaxHashStringSize) and should parse into
+// a chainhash.Hash without error.
+func validateTicketHash(c *gin.Context, hash string) (bool, error) {
+	if len(hash) != chainhash.MaxHashStringSize {
+		return false, fmt.Errorf("Incorrect hash length (clientIP=%s): got %d, expected %d", c.ClientIP(), len(hash), chainhash.MaxHashStringSize)
+
+	}
+	_, err := chainhash.NewHashFromStr(hash)
+	if err != nil {
+		return false, fmt.Errorf("Invalid hash (clientIP=%s): %v", c.ClientIP(), err)
```
jholdstock, line 162: helper functions constructing error messages with `clientIP` feels wrong. If you look at other helper funcs, none of them do this. They simply explain why they have failed, and then the caller constructs more detailed messages with the clientIP.

webapi/middleware.go
```diff
@@ -313,67 +303,36 @@ func vspAuth() gin.HandlerFunc {
 		// If the ticket was found in the database, we already know its
 		// commitment address. Otherwise we need to get it from the chain.
 		var commitmentAddress string
+		var isInvalid bool
```
jholdstock, line 306: This bool is only used inside the `else` branch, so it can be defined in there

webapi/middleware.go
```diff
@@ -313,67 +303,36 @@ func vspAuth() gin.HandlerFunc {
...
+				log.Errorf("%s: %v", funcName, err)
 				return
 			}
+		}
 
-			commitmentAddress = addr.String()
+		// Ensure a signature is provided.
+		signature := c.GetHeader("VSP-Client-Signature")
+		if signature == "" {
+			sendErrorWithMsg("no VSP-Client-Signature header", errBadRequest, c)
```
jholdstock, line 326: Log a warning with client IP if the client is sending bad requests.

jholdstock (changes requested):
I've left a few comments here. I think that addressing these comments will result in some significant changes to the PR, so to avoid wasting time I won't review in more detail until they are sorted.

jholdstock:
Going to merge this with a couple of minor outstanding issues. I will fix those myself in a follow-up PR shortly.

PR #331, opened 2022-03-18
gitignore .exe files

ukane-philemon:
This diff adds *.exe to list of gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.exe
 vspd
 /database/test.db
 /database/test.db-backup


.gitignore
```diff
@@ -1,3 +1,3 @@
-vspd
+*.exe
```
jholdstock, line 1: No issue with adding `*.exe`, but why remove `vspd`?

ukane-philemon: vspd don't generate folder `vspd`? I assumed it was intended to ignore `vspd.exe` file.

jholdstock: On Linux the executable is not named `vspd.exe`, is named just `vspd` without a file extension.

ukane-philemon: Thanks for your feedback, very helpful.

dajohi (approved):

Issue #339, opened 2022-03-28
Remove global vars.


jholdstock:
The vspd repo has so far only been used to build a single binary - `vspd`. Removing the use of global variables will make the code more reusable, which will make it possible to build multiple binaries, e.g. the recently introduced [`vote-validator`](https://github.com/decred/vspd/pull/335) and the standalone vspd testing tool [`v3tool`](https://github.com/dajohi/v3tool).

Work already started on this in #327.

PR #340, opened 2022-03-28
Remove global loggers.

jholdstock:
WIP - going to split some of this work into smaller independent PRs.

Part of #339


chappjc:
We went this route on the client side of dex, but left global loggers server side.  It solved some issues with concurrency and unit tests, but I often wish we could have kept global ones.  I'm curious what challenges motivated this for vspd, and if it was similar to dex client packages.

jholdstock:
#339 supplies a bit of additional context.

Removing global vars should make components easier to re-use because their dependencies will be explicitly stated (in their constructors) rather than having components silently depending on global state and then panicking when that global state is not initialised.

This is more important for some of the global variables in vspd than for others - eg. there are some bare global mutexes which I definitely want to take care of. Logging has a pretty well defined infrastructure around it so it is definitely lower priority, but I'd still like to take care of it if possible.

jholdstock:
Ready to go

database/database_test.go
```diff
@@ -82,15 +84,20 @@ func TestDatabase(t *testing.T) {
 	}
 
 	for testName, test := range tests {
+		// Set test logger to stdout.
+		backend := slog.NewBackend(os.Stdout)
+		log := backend.Logger("test")
+		log.SetLevel(slog.LevelTrace)
```
dnldd, line 90: Consider turning this into a test helper function (`newTestLogger` for example) to reuse. I'm also unsure why the logger is being created in a loop here. Consider creating it outside the loop.

dnldd (changes requested):
Looks good overall 👍

config.go
```diff
@@ -427,7 +437,7 @@ func loadConfig() (*config, error) {
 		}
 
 		// Create new database.
-		err = database.CreateNew(cfg.dbPath, cfg.FeeXPub)
+		err = database.CreateNew(cfg.dbPath, cfg.FeeXPub, cfg.Logger(" DB"))
```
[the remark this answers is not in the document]
jholdstock: It is. Making all of the subsystems 3 characters ensures they line up nicely in the log file. Its ugly and more difficult to read without the alignment.

database/database_test.go
jholdstock: Good points, will implement both. Thanks

jholdstock: The new func can't actually be reused without duplicating it because it is required in two separate packages. I am adding it anyway because its nice to have a func with a self explanatory name rather than a comment.

dnldd (approved):

PR #342, opened 2022-03-30
Move CanTicketVote out of dcrd RPC client.

jholdstock:
`CanTicketVote` doesn't really fit into the RPC client code as it is more of a business logic function. Moving it into the webapi package is more appropriate.

I suggested reviewing commits one-by-one.

Requesting review from @JoeGruffins because this touches a lot of the test code he wrote for `setaltsignaddr.go`


webapi/setaltsignaddr_test.go
```diff
@@ -167,107 +173,119 @@ func TestSetAltSignAddress(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			c.Request.Header.Set("VSP-Client-Signature", reqSig)
+
+			r.ServeHTTP(w, c.Request)
+
+			if test.wantCode != w.Code {
+				t.Errorf("expected status %d, got %d", test.wantCode, w.Code)
```
JoeGruffins, line 271: I guess I did this, but should be t.Fatalf, otherwise test may panic if this errors and test continues.

webapi/setaltsignaddr_test.go
```diff
@@ -167,107 +173,119 @@ func TestSetAltSignAddress(t *testing.T) {
...
+				if altsig != nil {
+					t.Fatalf("expected no alt sign addr saved for errored state")
+				}
+				return
+			}
+
+			if !bytes.Equal(b, []byte(altsig.Req)) || altsig.ReqSig != reqSig {
+				t.Fatalf("expected alt sign addr data different than actual")
+			}
+		})
```
JoeGruffins, line 289: nit: Can just be t.Fatal now because no name to format.

JoeGruffins (approved):
Looks good! Also tested buying on testnet, was fine.

PR #343, opened 2022-03-31
Miscellaneous housekeeping.

jholdstock:
Various small improvements broken out from #340 and #339. Split into small commits to ease review.

### DIFF
--- a/background/background.go
+++ b/background/background.go
@@ -72,9 +72,7 @@ func blockConnected() {
 
 	const funcName = "blockConnected"
 
-	ctx := context.Background()
-
-	dcrdClient, _, err := dcrdRPC.Client(ctx)
+	dcrdClient, _, err := dcrdRPC.Client()
 	if err != nil {
 		log.Errorf("%s: %v", funcName, err)
 		return
@@ -168,7 +166,7 @@ func blockConnected() {
 		log.Errorf("%s: db.GetUnconfirmedFees error: %v", funcName, err)
 	}
 
-	walletClients, failedConnections := walletRPC.Clients(ctx)
+	walletClients, failedConnections := walletRPC.Clients()
 	if len(walletClients) == 0 {
 		log.Errorf("%s: Could not connect to any wallets", funcName)
 		return
@@ -331,7 +329,7 @@ func blockConnected() {
 func connectNotifier(shutdownCtx context.Context, dcrdWithNotifs rpc.DcrdConnect) error {
 	notifierClosed = make(chan struct{})
 
-	dcrdClient, _, err := dcrdWithNotifs.Client(shutdownCtx)
+	dcrdClient, _, err := dcrdWithNotifs.Client()
 	if err != nil {
 		return err
 	}
@@ -423,15 +421,13 @@ func checkWalletConsistency() {
 
 	log.Info("Checking voting wallet consistency")
 
-	ctx := context.Background()
-
-	dcrdClient, _, err := dcrdRPC.Client(ctx)
+	dcrdClient, _, err := dcrdRPC.Client()
 	if err != nil {
 		log.Errorf("%s: %v", funcName, err)
 		return
 	}
 
-	walletClients, failedConnections := walletRPC.Clients(ctx)
+	walletClients, failedConnections := walletRPC.Clients()
 	if len(walletClients) == 0 {
 		log.Errorf("%s: Could not connect to any wallets", funcName)
 		return
--- a/database/database.go
+++ b/database/database.go
@@ -172,7 +172,8 @@ func CreateNew(dbFile, feeXPub string) error {
 
 // Open initializes and returns an open database. An error is returned if no
 // database file is found at the provided path.
-func Open(ctx context.Context, shutdownWg *sync.WaitGroup, dbFile string, backupInterval time.Duration, maxVoteChangeRecords int) (*VspDatabase, error) {
+func Open(shutdownCtx context.Context, shutdownWg *sync.WaitGroup, dbFile string,
+	backupInterval time.Duration, maxVoteChangeRecords int) (*VspDatabase, error) {
 
 	// Error if db file does not exist. This is needed because bolt.Open will
 	// silently create a new empty database if the file does not exist. A new
@@ -211,7 +212,7 @@ func Open(ctx context.Context, shutdownWg *sync.WaitGroup, dbFile string, backup
 				if err != nil {
 					log.Errorf("Failed to write database backup: %v", err)
 				}
-			case <-ctx.Done():
+			case <-shutdownCtx.Done():
 				shutdownWg.Done()
 				return
 			}
@@ -385,7 +386,7 @@ func (vdb *VspDatabase) BackupDB(w http.ResponseWriter) error {
 
 // CheckIntegrity will ensure that all data in the database is present and up to
 // date.
-func (vdb *VspDatabase) CheckIntegrity(ctx context.Context, dcrd rpc.DcrdConnect) error {
+func (vdb *VspDatabase) CheckIntegrity(dcrd rpc.DcrdConnect) error {
 
 	// Ensure all confirmed tickets have a purchase height.
 	// This is necessary because of an old bug which, in some circumstances,
@@ -400,7 +401,7 @@ func (vdb *VspDatabase) CheckIntegrity(ctx context.Context, dcrd rpc.DcrdConnect
 		return nil
 	}
 
-	dcrdClient, _, err := dcrd.Client(ctx)
+	dcrdClient, _, err := dcrd.Client()
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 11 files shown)


database/database.go
```diff
@@ -207,11 +208,11 @@ func Open(ctx context.Context, shutdownWg *sync.WaitGroup, dbFile string, backup
...
 			select {
 			case <-time.After(backupInterval):
-				err := writeHotBackupFile(db)
+				err := vdb.writeHotBackupFile()
 				if err != nil {
 					log.Errorf("Failed to write database backup: %v", err)
 				}
-			case <-ctx.Done():
+			case <-shutdownCtx.Done():
 				shutdownWg.Done()
```
JoeGruffins, line 216: I think it's very important to call `db.Close()` here before `shutdownWg.Done()`. I think the entire reason for having the waitgroup. We want to be sure the db is done writing before we shut down. If we cut it off mid write, that could lead to a corrupted db.

rpc/dcrd.go
```diff
@@ -58,7 +58,8 @@ func (d *DcrdConnect) Close() {
 
 // Client creates a new DcrdRPC client instance. Returns an error if dialing
 // dcrd fails or if dcrd is misconfigured.
-func (d *DcrdConnect) Client(ctx context.Context) (*DcrdRPC, string, error) {
+func (d *DcrdConnect) Client() (*DcrdRPC, string, error) {
+	ctx := context.Background()
```
JoeGruffins, line 62: What is the purpose of this context.

If it is to be sure that client can finish what it's doing and cleanly stop in a multi-threaded enviroment, then it needs to take a context and a waitgroup.

If it is to prevent long processes on shutdown from continuing then needs to take a context.

If it is to prevent it from taking to long, it should use `context.WithTimeout`

If it's just to make it work, I think that `context.WithCancel` is recommended, and you do `defer cancel` right after.

In my opinion, the ctx should come from higher up like it was. It *can* be saved in the struct (although rather all the methods just take a separate ctx). All of the rpc calls should take that ctx and further do `context.WithTimeout` with a reasonable timeout, like 5-10 seconds.

JoeGruffins, line 62: What is the purpose of this context.

If it is to be sure that client can finish what it's doing and cleanly stop in a multi-threaded enviroment, then it needs to take a context and a waitgroup.

If it is to prevent long processes on shutdown from continuing then needs to take a context.

If it is to prevent it from taking to long, it should use `context.WithTimeout`

If it's just to make it work, I think that `context.WithCancel` is recommended, and you do `defer cancel` right after.

In my opinion, the ctx should come from higher up like it was. It can be saved in the struct. All of the rpc calls should take that ctx and further do `context.WithTimeout` with a reasonable timeout, like 5-10 seconds.

webapi/middleware.go
```diff
@@ -71,7 +71,7 @@ func (s *Server) requireAdmin(c *gin.Context) {
 // downstream handlers to make use of.
 func withDcrdClient(dcrd rpc.DcrdConnect) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		client, hostname, err := dcrd.Client(c)
+		client, hostname, err := dcrd.Client()
```
JoeGruffins, line 74: I think this was good before. is `*gin.Context` doing something weird?

JoeGruffins, line 74: I think this was good before. Is there a reason for the change?

webapi/webapi.go
```diff
@@ -171,11 +171,11 @@ func Start(ctx context.Context, requestShutdown func(), shutdownWg *sync.WaitGro
...
-			case <-ctx.Done():
+			case <-shutdownCtx.Done():
 				shutdownWg.Done()
 				return
 			case <-time.After(refresh):
-				err := s.cache.update(ctx, vdb, dcrd, wallets)
+				err := s.cache.update(vdb, dcrd, wallets)
 				if err != nil {
 					log.Errorf("Failed to update cached VSP stats: %v", err)
 				}
```
JoeGruffins, line 181: I think the shutdown ctx was fine passed here.

JoeGruffins, line 181: I think the shutdown ctx was fine passed here.

background/background.go
```diff
@@ -423,15 +421,13 @@ func checkWalletConsistency() {
...
-
-	dcrdClient, _, err := dcrdRPC.Client(ctx)
+	dcrdClient, _, err := dcrdRPC.Client()
 	if err != nil {
 		log.Errorf("%s: %v", funcName, err)
 		return
 	}
 
-	walletClients, failedConnections := walletRPC.Clients(ctx)
+	walletClients, failedConnections := walletRPC.Clients()
```
JoeGruffins, line 430: Should use the shutdown ctx imo.

JoeGruffins, line 430: Should use the shutdown ctx imo.

rpc/dcrwallet.go
```diff
@@ -52,7 +52,8 @@ func (w *WalletConnect) Close() {
 // Clients loops over each wallet and tries to establish a connection. It
 // increments a count of failed connections if a connection cannot be
 // established, or if the wallet is misconfigured.
-func (w *WalletConnect) Clients(ctx context.Context) ([]*WalletRPC, []string) {
+func (w *WalletConnect) Clients() ([]*WalletRPC, []string) {
+	ctx := context.Background()
```
JoeGruffins, line 56: should take the context from higher up and do `context.WithTimeout` for rpc calls imo

JoeGruffins, line 56: should take the context from higher up and do `context.WithTimeout` for rpc calls imo

webapi/cache.go
```diff
@@ -56,8 +55,8 @@ func newCache(signPubKey string) *cache {
 
 // update will use the provided database and RPC connections to update the
 // dynamic values in the cache.
-func (c *cache) update(ctx context.Context, db *database.VspDatabase,
-	dcrd rpc.DcrdConnect, wallets rpc.WalletConnect) error {
+func (c *cache) update(db *database.VspDatabase, dcrd rpc.DcrdConnect,
+	wallets rpc.WalletConnect) error {
```
JoeGruffins, line 59: I think this was good before, taking the context.

JoeGruffins, line 59: I think this was good before, taking the context.

JoeGruffins (commented):
Srry for all the comments about the same thing.

I think the context could be used better in most of these methods. Comments inline.

JoeGruffins:
For some reason all of my comments got duplicated... Attempted to delete the duplicates.

database/database.go
JoeGruffins: OK! I see it now in `vspd.go`. Sorry I missed that. I think you want to still do `db.Close()` for the two possible errors in this function, when it would return with error and then that `.Close` will not happen in `vspd.go`

webapi/middleware.go
JoeGruffins: I see. Why do you need to connect to dcrd over and over, could just stay connected? Anyway, could run the context from the server I think.

rpc/dcrd.go
JoeGruffins: I updated my golangcli-lint and it does not complain here. So scratch that.

database/database.go
jholdstock: Good catch. I've pushed another commit for this.

webapi/middleware.go
jholdstock: `dcrd.Client()` makes use of [`rpc.client.dial()`](https://github.com/jholdstock/vspd/blob/d8dd02433c38eed51779fd84a46ac627b4ab14c8/rpc/client.go#L85-L109), which will return an existing connection if one exists. It only creates new connections if required.

rpc/dcrd.go
jholdstock: I looked a bit further at this and found that `TODO` and `Background` are identical:

https://github.com/golang/go/blob/2a6e13843d5bc0a380ce7081e33db9b636e394f9/src/context/context.go#L199-L218

I'd accept a PR changing from `Background` to `TODO`, f you feel strongly about this, but personally I don't think its worth the churn.

jholdstock:
I think the unresolved comments are all derivatives of the ones which have been resolved, so I think this PR is ready to merge @JoeGruffins ?

rpc/dcrd.go
JoeGruffins: I think the reason is having background in the middle of a function is misuse of context.

I could make an argument that this pr is `churn` without the change to TODO at least.

JoeGruffins (approved):
Tests well. But I'm still not happy about context use.

rpc/dcrd.go
jholdstock: Pushed one more change to go from Background to TODO in the rpc clients.

Upon re-reading these comments on a new day, it makes sense. Think I was just stuck in a weird thought pattern on Monday.

PR #359, opened 2022-11-21
Add client module

jholdstock:
Export a `client` module - a reusable HTTP client for vspd API consumers. This will be useful for dcrwallet, dcrwebapi, and testing tools such as [v3tool](https://github.com/dajohi/v3tool) which will [soon](https://github.com/decred/vspd/compare/master...jholdstock:vspd:v3tool2?expand=1) be moved into this repo.

The first commit on this PR is code copied directly from dcrwallet and the following commits are a few minor refactors.

@jrick - I'd appreciate a review from you on this because I'd like to see dcrwallet consume this module. I've already got a local diff for that which is working well.


client/client.go
```diff
@@ -227,6 +164,15 @@ func (c *Client) do(ctx context.Context, method, path string, addr stdaddr.Addre
...
 	}
 
+	type vspResponse struct {
+		Request []byte `json:"request"`
+	}
+
+	// Verify initial request matches server.
+	if !bytes.Equal(reqBody, resp.(vspResponse).Request) {
+		return fmt.Errorf("server response contains differing request")
+	}
```
jrick, line 174: currently in wallet, all of the POST methods do this check, but we do not perform it for GET /api/v3/vspinfo.  So this looks fine as long as that is tested and works.

client/client.go
```diff
@@ -0,0 +1,193 @@
...
+
+	respBody, err := io.ReadAll(reply.Body)
+	if err != nil {
+		return fmt.Errorf("read response body: %w", err)
+	}
+
+	status := reply.StatusCode
+
+	if status != http.StatusOK {
```
jrick, line 139: Is there a reason why wallet's handling of non-400 errors is different?  An issue we had run into before was where a reverse proxy was misconfigured, or vspd was not running/crashed/etc., and the 500 errors from the RP were difficult to distinguish from API errors.

jrick: or rather, we still wanted to parse the response body if vspd returned a 400 level error.

client/client.go
jholdstock: Good catch. I had thoroughly tested all of the POST funcs using v3tool, but not the GET one. I've pulled that logic out of do() up into post()

client/client.go
jholdstock: I'm not entirely sure what the original logic was, but afaict the errors returned by the new code will contain either the same or more information than those returned by the old code, so I think we are covered. I've added comments to explain the logic.

jrick: I think it's probably a little more resilient to decide how to decode the response or return an error based on the HTTP status code first, rather than reading the response body first, but I think this code should work for most situations and we should still see 500 level errors in the error message as long as the reverse proxy's response body is not a JSON object.

jholdstock: I don't think client should trust HTTP status for two reasons:

- Client will need to maintain its own list of statuses which it expects vspd to return, which is annoying and liable to fall out of date.
- Intermediary servers may return the same statuses as vspd returns, eg. RP might return 500 if vspd is down, vspd might return 500 if dcrd is down.

I just tried the "500 + JSON" scenario you described and I see that the code isnt working as I thought it would. I'll write some tests to make sure it works properly.

dajohi (approved):

client/client.go
jholdstock: Tests added in #363

client/client.go
jholdstock: Opened #364 to implement this.

PR #382, opened 2023-04-20
client: Automatic fee payment from dcrwallet

### LINKED ISSUE #379: multi: Add autoclient.
I believe there is an effort to move the rest of the `dcrwallet/internal/vsp` code here, so making an issue to track those changes.

Sample code from @jholdstock https://github.com/jholdstock/vspd/compare/dcr-deps...jholdstock:vspd:internal-vsp?expand=1

jholdstock:
Broken into commits to ease reviewing.

Closes #379


client/autoclient.go
```diff
@@ -0,0 +1,452 @@
...
+	"encoding/base64"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"sync"
+
+	"decred.org/dcrwallet/v3/errors"
+	"decred.org/dcrwallet/v3/wallet"
+	"decred.org/dcrwallet/v3/wallet/udb"
```
JoeGruffins, line 14: Could this and the feepayment file just use "decred.org/dcrwallet/v2" for now? v3 is not yet public, so it made sense to use when this was all internal, but now the consumer will have to replace for v3 modules in order to use the `Wallet` interface. I have confirmed v2 works still...

client/autoclient.go
```diff
@@ -0,0 +1,452 @@
...
+
+	// Dialer specifies an optional dialer when connecting to the VSP.
+	Dialer DialFunc
+
+	// Wallet specifies a loaded wallet.
+	Wallet Wallet
+
+	// Default policy for fee payments unless another is provided by the
+	// caller.
+	Policy Policy
```
JoeGruffins, line 91: nit: Can Policy be a pointer type? It doesn't make any functional difference really, just usually we would use a pointer here I think.

JoeGruffins (approved):
Have tried this out but replacing v3 wallet with v2 and it's working nicely. Thanks for making these changes this is a very useful package to have access to!

client/autoclient.go
```diff
@@ -0,0 +1,452 @@
...
+)
+
+type DialFunc func(ctx context.Context, network, addr string) (net.Conn, error)
+
+type Policy struct {
+	MaxFee     dcrutil.Amount
+	ChangeAcct uint32 // to derive fee addresses
+	FeeAcct    uint32 // to pay fees from, if inputs are not provided to Process
+}
+
```
JoeGruffins, line 31: Maybe it is best to insure that dcrwallet can satisfy this interface with the following somewhere:

`var _ WalletFetcher = (*wallet.Wallet)(nil)`

That way if a method changes in dcrwallet you will know when building here that this interface is no longer compatible.

client/autoclient.go
```diff
@@ -0,0 +1,452 @@
...
+type Wallet interface {
```
JoeGruffins, line 32: Maybe it is best to insure that dcrwallet can satisfy this interface with the following somewhere:

var _ Wallet = (*wallet.Wallet)(nil)

That way if a method changes in dcrwallet you will know when building here that this interface is no longer compatible.

jholdstock:
Cheers Joe, I've made all those changes.

client/feepayment.go
```diff
@@ -0,0 +1,962 @@
...
+		return nil
+	case "broadcast":
+		fp.log.Infof("VSP has successfully sent the fee tx for %v", &fp.ticketHash)
+		// Broadcasted, but not confirmed.
+		fp.schedule("confirm payment", fp.confirmPayment)
+		return nil
+	case "confirmed":
+		fp.remove("confirmed by VSP")
+		// nothing scheduled
+		err = w.UpdateVspTicketFeeToConfirmed(ctx, &fp.ticketHash, &fp.feeHash, fp.client.URL, fp.client.PubKey)
```
buck54321, line 938: I think you need a
```Go
fp.mu.Lock()
feeHash := &fp.feeHash
fp.mu.Unlock()
err = w.UpdateVspTicketFeeToConfirmed(ctx, &fp.ticketHash, feeHash, fp.client.URL, fp.client.PubKey)
```

client/feepayment.go
```diff
@@ -0,0 +1,962 @@
...
+		var apiErr types.ErrorResponse
+		if errors.As(err, &apiErr) && apiErr.Code == types.ErrFeeExpired {
+			// Fee has been expired, so abandon current feetx, set fp.feeTx
+			// to nil and retry submit payment to make a new fee tx.
+			feeHash := feeTx.TxHash()
+			err := w.AbandonTransaction(ctx, &feeHash)
+			if err != nil {
+				fp.log.Errorf("error abandoning expired fee tx %v", err)
+			}
+			fp.feeTx = nil
```
buck54321, line 865:
```Go
fp.mu.Lock()
fp.feeTx = nil
fp.mu.Unlock()
```

client/feepayment.go
```diff
@@ -0,0 +1,962 @@
...
+	"github.com/decred/dcrd/dcrutil/v4"
+	"github.com/decred/dcrd/txscript/v4"
+	"github.com/decred/dcrd/txscript/v4/stdaddr"
+	"github.com/decred/dcrd/txscript/v4/stdscript"
+	"github.com/decred/dcrd/wire"
+	"github.com/decred/slog"
+	"github.com/decred/vspd/types/v2"
+)
+
+var prng lockedRand
```
buck54321, line 34: Could forego the `init` function and just 
```Go
var prng = lockedRand{
	rand: rand.New(rand.NewSource(time.Now().UnixNano())),
}
```

jholdstock:
Thanks Buck. I found another mutex issue in `TrackedTickets()`, fixed it in the same commit.

jrick:
Replacing crypto/rand with math/rand doesn't sit well with me.

You don't need a PRNG here, so the calls to uniformprng could be replaced with using crypto/rand directly.

For the random jitter duration, I would just copy the body of the uniformprng Int63n method, but read a random uint64 from crypto/rand.

jholdstock:
Rebased and now using `crypto/rand` per a conversation with jrick.

@buck54321 @JoeGruffins please test these latest changes against your code and see if this satisfies your requirements.

JoeGruffins:
Tested at `0ec0813` and seems to be working good.

jholdstock:
I believe all feedback has been addressed and this can be merged.

I pushed one last commit to move this back to dcrwallet/v3 in order to get this working against master dcrd/dcrwallet.

PR #384, opened 2023-06-04
build: Update linter to 1.53.2

jholdstock:


database/encoding_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 The Decred developers
+// Copyright (c) 2022-2023 The Decred developers
```
davecgh, line 1: Doesn't look like anything changed in this file to warrant this bump.

dajohi (approved):

PR #387, opened 2023-06-09
client: Use latest dcrd deps.

davecgh:
This updates the client module to remove all replace directives and use the latest release versions of the dcrd deps.


davecgh:
> looks good but i am hesitant to get wallet using this yet. so maybe no need for an immediate merge and/or tag.

Yeah, I don't plan to tag it.  I think we're still want to update to use the wallet/v3@v3.0.0 release module first before tagging a new client too.

PR #389, opened 2023-06-11
main: Use latest dcrd deps.

jholdstock:


davecgh (approved):

go.mod
```diff
@@ -27,23 +27,17 @@ require (
 	go.etcd.io/bbolt v1.3.7
 )
 
-replace (
```
davecgh, line 30: Just an FYI.  In the future if you want to consume development modules from another repository, like in this case, instead of using replaces, you should specifically require the minimum necessarily in-development commits each one needs.  That way it doesn't break consumers of your own module.

davecgh: For example instead of requiring `github.com/decred/dcrd/blockchain/stake/v5 v5.0.0` and then replacing `github.com/decred/dcrd/blockchain/stake/v5` with `github.com/decred/dcrd/blockchain/stake/v5 v5.0.0-20221022042529-0a0cc3b3bf92`, instead just require `github.com/decred/dcrd/blockchain/stake/v5 v5.0.0-20221022042529-0a0cc3b3bf92` directly.

jholdstock: Thanks for the tip!  Very helpful, it actually explains a couple of issues I've seen recently.

PR #395, opened 2023-06-25
client: Remove all remaining wallet dependencies

jholdstock:
Incomplete WIP, do not review yet.


go.mod
```diff
@@ -27,10 +27,15 @@ require (
 	go.etcd.io/bbolt v1.3.7
 )
 
+replace decred.org/dcrwallet/v4 => ../dcrwallet
```
davecgh, line 30: This shouldn't be committed as it doesn't work for consumers.  If you want it for local purposes, you should put it in your local `go.work`.

jholdstock:
Cancelling the autoclient effort. #403 removes this code.

PR #398, opened 2023-07-01
rpc: Ignore another "duplicate tx" error.

jholdstock:
sendrawtransaction is expected to fail if the transaction already exists in the mempool or a mined block. dcrd returns two different errors which can indicate this situation. Handling for one of them already existed, and this commit adds handling for the other.


davecgh (approved):
This looks accurate and should address the issue.

dajohi (approved):

Issue #268, opened 2021-06-05
Add missed and expired ticket count to vspinfo response


xaur:
`api/v3/vspinfo` only returns `revoked` ticket count. dcrstakepool's `api/v2/stats` reports `Expired`, `Missed`, and `Expired` separately, allowing to determine VSP's miss ratio with "unlucky" expired tickets subtracted. Miss ratio is an important indicator when choosing a VSP.

I understand the operators can cheat and patch the code to lie about it, but that would carry its own reputational risks. Adding `missed` and `expired` to the API would be helpful.

jholdstock:
Agreed this would be nice to have, but its not actually possible with the information exposed by the voting wallets RPC. The best we can do is detect if a ticket is revoked or not.

I can't remember the exact reason for this, but it might have been related to the fact that the tickets don't actually belong to the voting wallet - the purchase tx and the voting priv key have been imported from another wallet.

Quote from #86

> Using ticketinfo allows vspd to determine when a ticket is no longer votable, either because it has already voted or because it has been revoked. This information is now stored in the vspd database. The ticketinfo RPC does not distinguish between missed and expired tickets, and so vspd simply buckets both of these into revoked.

xaur:
Thanks for the info. Just to be clear, is there no way to derive from the blockchain if the ticket is missed or expired?

jholdstock:
The information can be retrieved from the chain. A recent example randomly pulled from dcrdata...

Revocation:
https://dcrdata.decred.org/tx/454a852eb302d541571062bb888330d6c46242b6e4934c307fb30643b07e656a

Linking back to its original ticket (status == missed):
https://dcrdata.decred.org/tx/00ea86def0d5a68cecae0e20e3dcb1c3b1ffb1a732403ceb0912806f57b9ffde


I may be incorrect about this, but I don't think SPV wallets can distinguish between missed and expired (although thats irrelevant for vspd, it never uses SPV)

xaur:
Oh good to know the missed|expired outcome can be reconstructed.

Also I wouldn't assume that all VSP users are SPV. Syncing dcrd on a PC is not that hard, running a reliable voting wallet 24/7 is much harder.

jholdstock:
I mean that vspd voting wallets are never SPV. Not talking about users.

jholdstock:
Seeing as this has been requested multiple times, reopening the issue to either:

- get a definitive answer for why it cannot be done
[...]

PR #406, opened 2023-08-18
Hoist database integrity checks up to vspd.

jholdstock:
This lays the groundwork for adding new integrity checks in future. The first of which will be implementing #268, using gcs filters to update all tickets in `revoked` status to either be `missed` or `expired`.

Requires #408 and #409


davecgh (commented):
Looks good once #409 is updated and this is rebased accordingly.  Notably, the context would change.

jholdstock:
Merging this without the context bits. Will address that separately because its a larger issue.

PR #408, opened 2023-08-22
Various refactoring

jholdstock:
A few minor improvements made whilst working on a larger project.

These are easy to review individually and add value on their own, so I've broken them out into their own PR.

### DIFF
--- a/database/ticket.go
+++ b/database/ticket.go
@@ -91,6 +91,25 @@ type Ticket struct {
 	Outcome TicketOutcome
 }
 
+type TicketList []Ticket
+
+// EarliestPurchaseHeight returns the lowest non-zero purchase height in the
+// list of tickets. Zero will be returned if the list is empty, or if every
+// ticket in the list has zero purchase height.
+func (t TicketList) EarliestPurchaseHeight() int64 {
+	var oldestHeight int64
+	for _, ticket := range t {
+		// Skip unconfirmed tickets.
+		if ticket.PurchaseHeight == 0 {
+			continue
+		}
+		if oldestHeight == 0 || oldestHeight > ticket.PurchaseHeight {
+			oldestHeight = ticket.PurchaseHeight
+		}
+	}
+	return oldestHeight
+}
+
 func (t *Ticket) FeeExpired() bool {
 	now := time.Now()
 	return now.After(time.Unix(t.FeeExpiration, 0))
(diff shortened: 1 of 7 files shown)


database/ticket.go
```diff
@@ -91,6 +91,25 @@ type Ticket struct {
 	Outcome TicketOutcome
 }
 
+type TicketList []Ticket
```
davecgh, line 94: This is nothing technically incorrect about this, but I'm curious if this is a precursor to something larger?

I ask because I've personally never liked it when code does this in general without a specific need since it hides the type for no real reason.  For example, when I'm looking at the code and am not already familiar with it and I see:

```Go
func (vdb *VspDatabase) GetUnconfirmedTickets() (TicketList, error) {
```

I think, hmm, ok, now I have to go figure out what `TicketList` is.  Is it a map?  A slice?  A generic container? Something else?

In other words, I generally find that it adds more cognitive overhead in general.

rpc/dcrd.go
```diff
@@ -228,9 +228,19 @@ func (c *DcrdRPC) GetBestBlockHeader() (*dcrdtypes.GetBlockHeaderVerboseResult,
...
+	if err != nil {
+		return nil, err
+	}
+	return blockHeader, nil
+}
+
+// GetBlockHeaderVerbose uses getblockheader RPC with verbose=true to retrieve
+// the header of the requested block.
+func (c *DcrdRPC) GetBlockHeaderVerbose(blockHash string) (*dcrdtypes.GetBlockHeaderVerboseResult, error) {
 	verbose := true
```
davecgh, line 241: No need for a local variable here.

Suggest:

```Go
const verbose = true
```

dajohi (approved):

database/ticket.go
jholdstock: Interesting point I hadn't considered before. I do anticipate adding more funcs to TicketList later, but nothing imminent. 

It's mainly just to make the code calling `EarliestPurchaseHeight` cleaner, but also has the side-effect of making autocomplete more useful. Probably a habit from my (mostly) OOP background.

rpc/dcrd.go
jholdstock: Good call, theres a few more of those to update as well.

davecgh (approved):

database/ticket.go
davecgh: Gotcha.  Like I said, nothing wrong with it technically.  I just figured I'd offer my perspective since I was reviewing.

PR #409, opened 2023-08-22
vspd: Introduce vspd struct.

jholdstock:
Storing all of the essential resources such as db, rpcs and logger in a vspd struct enables vspd functions to be called without passing a bunch of parameters unnecessarily. This will be increasingly useful later when further changes are introduced.

Requires #408 and #410


cmd/vspd/main.go
```diff
@@ -28,122 +30,143 @@ const maxVoteChangeRecords = 10
...
+	dcrd        rpc.DcrdConnect
+	wallets     rpc.WalletConnect
+}
 
-	// Load config file and parse CLI args.
-	cfg, err := loadConfig()
+// newVspd creates the essential resources required by vspd - a database, logger
+// and RPC clients - then returns an instance of vspd which is ready to be run.
+func newVspd(cfg *config) (*vspd, error) {
+
```
davecgh: Extra blank line. I'm kind of surprised the linter isn't complaining.  I saw several other cases of this, but only commenting once.

cmd/vspd/main.go
```diff
@@ -28,122 +30,143 @@ const maxVoteChangeRecords = 10
...
-	// Open database.
-	db, err := database.Open(cfg.dbPath, dbLog, maxVoteChangeRecords)
-	if err != nil {
-		log.Errorf("Database error: %v", err)
-		return 1
+// run starts all of vspds background services including the web server, and
+// stops all started services when a shutdown is requested.
+func (v *vspd) run() int {
+
+	v.log.Criticalf("Version %s (Go version %s %s/%s)", version.String(), runtime.Version(),
```
davecgh: Critical is supposed to be for fatal failures.  Most libs even do full on aborts `os.Exit` style on critical.  For example, [see how wallet only uses it with the Fatal family of funcs when implementing the stdlib interface](https://github.com/decred/dcrwallet/blob/a87fa843495ec57c1d3b478c2ceb3876c3749af5/internal/rpc/rpcserver/log.go#L56).

I know the code was already this way, but it's not really correct.

cmd/vspd/main.go
```diff
@@ -28,122 +30,143 @@ const maxVoteChangeRecords = 10
...
+		v.log.Warnf("")
+		v.log.Warnf("\tWARNING: Config --vspclosed is set. This will prevent vspd from accepting new tickets.")
+		v.log.Warnf("")
+	}
 
-	// Create a context that is cancelled when a shutdown request is received
-	// through an interrupt signal.
-	shutdownCtx := shutdownListener(log)
+	// Defer shutdown tasks.
+	defer v.log.Criticalf("Shutdown complete")
```
davecgh: Similarly, this shouldn't be critical.

cmd/vspd/main.go
```diff
@@ -28,122 +30,143 @@ const maxVoteChangeRecords = 10
...
 
-// run is the main startup and teardown logic performed by the main package. It
-// is responsible for parsing the config, creating dcrd and dcrwallet RPC clients,
-// opening the database, starting the webserver, and stopping all started
-// services when a shutdown is requested.
-func run() int {
+type vspd struct {
+	cfg         *config
+	log         slog.Logger
+	shutdownCtx context.Context
```
davecgh: Contexts shouldn't be stored in structs.   

See the [stdlib context doco](https://pkg.go.dev/context):

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

dajohi (approved):

cmd/vspd/main.go
jholdstock: Fixed

cmd/vspd/main.go
jholdstock: Fixed

cmd/vspd/main.go
jholdstock: Commented above

PR #410, opened 2023-08-22
vspd: Simplify shutdown listener.

jholdstock:
The existing signal.go was inherited directly from dcrwallet, and was slightly more flexible/complicated than required for vspd. This brings the code more inline with the dcrd implementation which can be instantiated in one line and with only one parameter.


ukane-philemon (approved):
Looks good.

PR #411, opened 2023-08-23
Extra linters to address common review points.

jholdstock:
Some of these cover points which have recently been highlighted in reviews, such as contexts contained in structs, slices initialized with zero length, and periods missing from the end of comments.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,6 +17,6 @@ jobs:
       - name: Build
         run: go build ./...
       - name: Install Linters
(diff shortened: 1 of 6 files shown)


davecgh (approved):
Looks good overall.

I understand this is focusing on removing the contexts from the struct, so nothing to hold the PR up over, but to offer a suggestion for future PRs, pretty much all of the cases that are wrappers for the RPCs (e.g. `GetRawTransaction`) should really be taking contexts so that each individual call can be independently cancelled and traced (if desired) as that is kind of the entire purpose of contexts in the RPCs.

Generally speaking, the callers to those functions would create a new context that wraps the overall shutdown context (or the already wrapped context for a subsystem, etc) to ensure the they are cancelled on shutdown due to context cancellation propagation.

davecgh (commented):
Reviewed the latest commit to add more linters too.

PR #412, opened 2023-08-23
rpc: Fix block connected handler.

jholdstock:
Fix a bug (introduced in #407) where it appeared as though a notification handler was being attached to the dcrd RPC client, but actually no notifications were received until the client was disconnected a new one was created.

Now, rather than delaying attaching the notification handler, it is attached immediately upon client creation and vspd does not start handling the received notifications until it is ready.

### DIFF
--- a/cmd/vspd/vspd.go
+++ b/cmd/vspd/vspd.go
@@ -42,6 +42,8 @@ type vspd struct {
 	db      *database.VspDatabase
 	dcrd    rpc.DcrdConnect
 	wallets rpc.WalletConnect
+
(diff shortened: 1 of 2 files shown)


davecgh (approved):
Nice catch. I missed this on my review of 407 as well.  I wouldn't have expected the RPC client to only register for `notifyblocks` on connection while simultaneously offering the ability to dynamically attach a handler.

Given that client behavior, it makes much more sense to require the creation of a channel during setup and nuke the ability to asynchronously attach a handler as this does.

PR #413, opened 2023-08-24
multi: Find voted/revoked tickets with GCS filters

jholdstock:
Use dcrd and GCS filters to find voted/revoked tickets rather than using the dcrwallet TicketInfo RPC.

Using TicketInfo was a bit flakey because wallets do not always correctly detect votes/revokes, and as a result VSP admins may notice that with this change vspd detects some historic voted/revoked tickets which TicketInfo never detected.


cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+	for i := range tx.Vin {
+		input := &tx.Vin[i]
+		if input.Vout == 0 && input.Txid == outputHash {
+			return true // Found spender.
+		}
+	}
+	return false
+}
+
+func pluralize(n int, s string) string {
```
davecgh, line 158: This works for the only usage there currently is, but I'd suggest stealing [pickNoun](https://github.com/decred/dcrd/blob/0a35adc02120805f5185cc0a53eb2674711e3013/log.go#L179-L186) for this purpose because not all words can be made plural by just adding an s.

cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+// height of the most recent scanned block.
+func (v *vspd) findSpentTickets(toCheck database.TicketList, startHeight int64) (map[string]spentTicket, int64, error) {
+	params := v.cfg.netParams
+
+	dcrdClient, _, err := v.dcrd.Client()
+	if err != nil {
+		return nil, 0, err
+	}
+
+	bestBlock, err := dcrdClient.GetBestBlockHeader()
```
davecgh, line 42: This isn't really performance critical, so it's not a big deal, but given the entire point of this is to find the most recent height, could just call [getblockcount](https://github.com/decred/dcrd/blob/0a35adc02120805f5185cc0a53eb2674711e3013/docs/json_rpc_api.mediawiki#user-content-getblockcount) rather than getting a header and decoding it.

cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+			break
+		}
+	}
+
+	return spent, endHeight, nil
+}
+
+// txSpendsOutput returns true if the passed tx has an input that spends the
+// specified output.
+func txSpendsOutput(tx types.TxRawResult, outputHash string) bool {
```
davecgh, line 148:
```suggestion
func txSpendsTicket(tx types.TxRawResult, outputHash string) bool {
```

Reading the call site before digging into the func, I originally suspected it was doing what the func name claim which wouldn't have been correct, but I see the func actually does check output zero.  The suggested name would make it clearer imo.

cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+
+	spent := make(map[string]spentTicket)
+
+	for iHeight := startHeight; iHeight <= endHeight; iHeight++ {
+		iHash, err := dcrdClient.GetBlockHash(iHeight)
+		if err != nil {
+			return nil, 0, err
+		}
+
+		iBlock, err := dcrdClient.GetBlockVerbose(iHash)
```
davecgh, line 89: I imagine you're not super concerned with performance here, but it would be a lot more efficient to avoid getting verbose results and instead just get the block itself (by setting the two verbose flags in the RPC).

With both flags false, you just get the raw hex-encoded block.  So you take that and decode it in a `wire.MsgBlock`.  I know you're not using dcrd's RPC client, but relevant code [here](https://github.com/decred/dcrd/blob/0a35adc02120805f5185cc0a53eb2674711e3013/rpcclient/chain.go#L73-L92).

That way you wouldn't need to round trip through additional hex encoding nor incur the overhead the extra things verbose provides that you don't need here.

cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+	heightSpent  int64
+	spendingTx   *wire.MsgTx
+}
+
+func (s *spentTicket) voted() bool {
+	return stake.IsSSGen(s.spendingTx)
+}
+
+// findSpentTickets attempts to find transactions that vote/revoke the provided
+// tickets by matching the payment script of the tickets commitment address
```
davecgh, line 30:
```suggestion
// tickets by matching the payment script of the ticket's commitment address
```

rpc/dcrd.go
```diff
@@ -261,3 +264,51 @@ func (c *DcrdRPC) ExistsLiveTicket(ticketHash string) (bool, error) {
...
+	if err != nil {
+		return key, nil, fmt.Errorf("getcfilterv2 error: %w", err)
+	}
+
+	filterB, err := hex.DecodeString(resp.Data)
+	if err != nil {
+		return key, nil, fmt.Errorf("error decoding block filter: %w", err)
+	}
+
+	filter, err := gcs.FromBytesV2(blockcf2.B, blockcf2.M, filterB)
```
davecgh, line 303: Since this is happening via RPC and the VSP probably controls the dcrd that is providing the filters, it isn't a huge concern, but this should ideally check the inclusion proof as well once the header commitments are active (e.g. block `000000000000000010815bed2c4dc431c34a859f4fc70774223dde788e95a01e` (height 431488) on mainnet).

e.g.
```Go
func (c *DcrdRPC) GetCFilterV2(blockHash string, header *wire.BlockHeader, verifyProof bool) ([gcs.KeySize]byte, *gcs.FilterV2, error) {
	...

	if verifyProof {
		filterHash := filter.Hash()
		if !standalone.VerifyInclusionProof(&header.StakeRoot, &filterHash, resp.ProofIndex, resp.ProofHashes) {
			/* error */
		}
	}

	return blockcf2.Key(&header.MerkleRoot), filter, nil
}
```

cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+		key, filter, err := dcrdClient.GetCFilterV2(iBlock.Hash, iBlock.MerkleRoot)
+		if err != nil {
+			return nil, 0, err
+		}
+
+	outer:
+		for ticketHash, ticket := range tickets {
+			if filter.Match(key, ticket.pkScript) {
+				// Ticket is likely spent in block. Confirm.
+				for _, blkTx := range append(iBlock.RawTx, iBlock.RawSTx...) {
```
davecgh, line 103: Tickets can only be spent by transactions in the stake tree (namely a vote or a revocation).  There isn't any need to range over the regular txns.

cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+					if !txSpendsOutput(blkTx, ticketHash) {
+						continue
+					}
+
+					// Confirmed - ticket is spent in block.
+
+					decoded, err := hex.DecodeString(blkTx.Hex)
```
davecgh, line 110: If you decode the raw block as suggested above, this wouldn't be necessary because you would be ranging over the already decoded stake txns.

cmd/vspd/spentticket.go
```diff
@@ -0,0 +1,164 @@
...
+					if err != nil {
+						return nil, 0, fmt.Errorf("could not decode hex %q: %w", blkTx.Hex, err)
+					}
+
+					msgTx := &wire.MsgTx{}
```
davecgh, line 115: Nor this.

cmd/vspd/spentticket.go
jholdstock: I'm going to leave this as is because I personally prefer the simplicity at the call sites. It will be enough to update later if required. Even in dcrd which uses pickNoun a whole bunch I notice that it never actually does anything more than add an "s" at the moment.

I've added a comment to make the limitation really explicit.

jholdstock:
Thanks @davecgh. Performance isn't critical here, but its great to have your suggestions because there are (rare) cases where vspd might need to rescan several thousand blocks. I added some temporary logs for timing and the improvement is notable.

**Before**

```no-highlight
[DBG] VSP: Scanning 46552 blocks for 477 spent tickets
[DBG] VSP: took 2m10.727347019s
```

**After**

```no-highlight
[DBG] VSP: Scanning 46554 blocks for 477 spent tickets
[DBG] VSP: took 28.316309518s
```

This PR now contains three commits - the first is a new one to remove simnet params because they are completely unused/untested, and I was too lazy to figure out how to handle the DCP-0005 activation height. The middle commit contains exactly the same changes you already reviewed, and the third is a new one to implement your suggestions.

davecgh (commented):
Nice.  Glad to hear the suggestions helped.

davecgh:
> ... I was too lazy to figure out how to handle the DCP-0005 activation height.

That one is easy.  It's always active on simnet.

davecgh (approved):

davecgh:
Nice job overall and it look good now.  It much more robust than what it's replacing and with the latest updates should stay reasonably efficient.

As an aside, if you ultimately end up need more performance, you could use [filter.MatchAny](https://pkg.go.dev/github.com/decred/dcrd/gcs/v4#FilterV2.MatchAny) instead because it saves a lot of time on the filter matching itself when looking for multiple things like this.  To do that though, you'd need to restructure you code slightly in this section to use slices instead of a map.

PR #415, opened 2023-08-26
multi: Update revoked tickets to expired/missed.

jholdstock:
Any tickets in the database which are currently revoked should be updated to either expired or missed.

This is achieved with a heuristic based on the expiry height and revoke height of the tickets. It is not guaranteed to be 100% correct but should be pretty close.

The web api is not yet updated to reflect this change, missed/expired tickets will continue to be counted as revoked.

### DIFF
--- a/cmd/vspd/vspd.go
+++ b/cmd/vspd/vspd.go
@@ -219,7 +219,12 @@ func (v *vspd) run() int {
 func (v *vspd) checkDatabaseIntegrity() error {
 	err := v.checkPurchaseHeights()
 	if err != nil {
-		return err
+		return fmt.Errorf("checkPurchaseHeights error: %w", err)
+	}
+
+	err = v.checkRevoked()
+	if err != nil {
+		return fmt.Errorf("checkRevoked error: %w", err)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


cmd/vspd/vspd.go
```diff
@@ -270,6 +275,62 @@ func (v *vspd) checkPurchaseHeights() error {
...
+	if err != nil {
+		return fmt.Errorf("db.GetRevoked error: %w", err)
+	}
+
+	if len(revoked) == 0 {
+		// Nothing to do, return.
+		return nil
+	}
+
+	v.log.Warnf("Updating %s in revoked status, this may take a while...",
```
jholdstock, line 291: Adding context for "this may take a while", on my testnet VSP 

```no-highlight
[WRN] VSP: Updating 231 tickets in revoked status, this may take a while...
[DBG] VSP: Scanning 109113 blocks for 231 spent tickets
[INF] VSP: Took 36.742324728s
```

davecgh (approved):
Untested, but I don't see any issues.  The heuristic is as we previously discussed and the upgrade logic looks correct.

PR #416, opened 2023-08-29
vspd: Use filter.MatchAny instead of filter.Match.

jholdstock:
This brings a notable performance improvement to spent ticket scanning without changing functionality.

### DIFF
--- a/cmd/vspd/spentticket.go
+++ b/cmd/vspd/spentticket.go
@@ -170,7 +170,7 @@ func (v *vspd) findSpentTickets(toCheck database.TicketList, startHeight int64) 
 				scripts = nonOrderPreservingRemove(scripts, i)
 
 				// Current index has been removed which means everything else
-				// moved up one and thus the same index needed to be repeated.
+				// moved up one and thus the same index needs to be repeated.
 				i--
 			}
 		}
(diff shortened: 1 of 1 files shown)


davecgh:
> This brings a notable performance improvement to spent ticket scanning without changing functionality.

To quantify, it cuts the scanning time roughly in half.

cmd/vspd/spentticket.go
```diff
@@ -133,42 +134,40 @@ func (v *vspd) findSpentTickets(toCheck database.TicketList, startHeight int64)
...
+					dbTicket:     ticket.dbTicket,
+					expiryHeight: ticket.dbTicket.PurchaseHeight + int64(params.TicketMaturity) + int64(params.TicketExpiry),
+					heightSpent:  iHeight,
+					spendingTx:   blkTx,
+				})
+
+				// Remove this ticket and it's script before continuing with the
+				// next one.
+				tickets = nonOrderPreservingRemove(tickets, i)
+				scripts = nonOrderPreservingRemove(scripts, i)
```
davecgh, line 170: Don't forget to do `i--`.  You removed the current index which means everything else moved up one and thus the same index needed to be repeated.

davecgh (changes requested):

PR #417, opened 2023-08-29
Display expired/missed tickets on webpage

jholdstock:
In order to make space for the new counters the network is no longer displayed in the stats row. It is replaced with a banner which is only shown on testnet. PR also includes a fix for a minor bug I noticed along the way.

Screenshot from 2023-08-29 12-34-44

Contributes to #268

### DIFF
--- a/webapi/cache.go
+++ b/webapi/cache.go
@@ -108,16 +108,13 @@ func (c *cache) update(db *database.VspDatabase, dcrd rpc.DcrdConnect,
 	c.data.BlockHeight = bestBlock.Height
 	c.data.NetworkProportion = float32(voting) / float32(bestBlock.PoolSize)
 
-	// Prevent dividing by zero when pool has no voted tickets.
-	switch voted {
-	case 0:
-		if revoked == 0 {
-			c.data.RevokedProportion = 0
-		} else {
-			c.data.RevokedProportion = 1
-		}
-	default:
-		c.data.RevokedProportion = float32(revoked) / float32(voted)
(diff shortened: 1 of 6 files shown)


webapi/cache.go
```diff
@@ -104,20 +106,29 @@ func (c *cache) update(db *database.VspDatabase, dcrd rpc.DcrdConnect,
...
+	total := voted + expired + missed
+
+	// Prevent dividing by zero when pool has no voted/expired/missed tickets.
+	switch total {
 	case 0:
-		if revoked == 0 {
-			c.data.RevokedProportion = 0
+		if expired == 0 {
+			c.data.ExpiredProportion = 0
+		} else {
```
davecgh, line 121: These else branches look impossible to me and can be removed imo.  This is under `case 0`, meaning `total` is `0` and `total := voted + expired + missed`.  So, there is no way `expired` can be non-zero.

jholdstock: Good spot, updated the existing commits with fixes.

webapi/cache.go
```diff
@@ -104,20 +106,29 @@ func (c *cache) update(db *database.VspDatabase, dcrd rpc.DcrdConnect,
...
+			c.data.ExpiredProportion = 1
+		}
+		if missed == 0 {
+			c.data.MissedProportion = 0
```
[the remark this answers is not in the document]
jholdstock: Fixed

davecgh (approved):

PR #420, opened 2023-08-31
Add expired/missed tickets to /vspinfo

jholdstock:
Adding fields for expired/missed to /vspinfo response, deprecate revoked.

This will need a minor version bump in the types module.


davecgh (approved):
Changes look fine.  Not sure if people use VSP APIs, but if so, this is changing an exported type (adding to it), so the module minor version should get a bump (if it hasn't already had one since last release).

PR #421, opened 2023-08-31
webapi: Add expired/missed tickets to /vspinfo

### LINKED ISSUE #268: Add missed and expired ticket count to vspinfo response

jholdstock:
Requires #417 
Closes #268


davecgh (approved):
Looks good.  Will need rebase after #417 is updated before merge.

PR #424, opened 2023-09-05
Consolidate background task timers.

jholdstock:
Using a single select loop for background tasks removes a lot of duplicated boilerplate code and helps to simplify shutdown logic. This does reduce the amount of things which can run in parallel, but that isn't of concern for vspd. The web server still runs in its own goroutine so its responsiveness won't be affected.


jholdstock:
Included in #426

PR #428, opened 2023-09-11
Move webapi and version packages to internal.

jholdstock:
These two packages are almost entirely just for the internal implementation of vspd and are not really suitable for exporting or consuming by third parties.


internal/webapi/setaltsignaddr.go
```diff
@@ -17,21 +17,21 @@ import (
 )
 
 // Ensure that Node is satisfied by *rpc.DcrdRPC.
-var _ Node = (*rpc.DcrdRPC)(nil)
+var _ node = (*rpc.DcrdRPC)(nil)
 
-// Node is satisfied by *rpc.DcrdRPC and retrieves data from the blockchain.
-type Node interface {
+// node is satisfied by *rpc.DcrdRPC and retrieves data from the blockchain.
+type node interface {
```
davecgh, line 23: I double checked that this isn't used outside of the package.  It's only used as a parameter to `canTicketVote` so it can be mocked in tests.  Thus, I agree it's fine to be unexported.

PR #429, opened 2023-09-14
vspd: Copy VSP fee validation func from wallet.

jholdstock:
This is such a trivial func that there is no need to import it from wallet, it can just be copied into this project.


davecgh (approved):
Confirmed the function matches.

PR #430, opened 2023-09-14
Move netparams to new internal package.

jholdstock:
A new internal package named config contains all of the hard-coded, network specific values used by vspd.


internal/webapi/getfeeaddress.go
```diff
@@ -62,7 +62,7 @@ func (s *server) getCurrentFee(dcrdClient *rpc.DcrdRPC) (dcrutil.Amount, error)
 	}
 
 	fee := txrules.StakePoolTicketFee(sDiff, defaultMinRelayTxFee,
-		int32(bestBlock.Height), s.cfg.VSPFee, s.cfg.NetParams, isDCP0010Active)
+		int32(bestBlock.Height), s.cfg.VSPFee, s.cfg.Network.Params, isDCP0010Active)
```
davecgh, line 65: Not sure how strict you are in vspd repo, but this and several others are > 80 now.  I see others that were already that way too, hence I'll avoid further commenting on it.

internal/config/network.go
```diff
@@ -0,0 +1,61 @@
...
+	// MinWallets is the minimum number of voting wallets required for a vspd
+	// deployment on this network. vspd will log an error and refuse to start if
+	// fewer wallets are configured.
+	MinWallets int
+	// DCP0005Height is the activation height of DCP-0005 block header
+	// commitments agenda on this network.
+	DCP0005Height int64
+}
+
+var MainNet = &Network{
```
davecgh, line 25: Probably not a big deal in vspd, but making these pointers means they have to be scanned on every single GC pass for the entire life of the process which eats extra CPU time.  They should be concrete structs for best performance and then only have their addresses taken at the call sites.

internal/config/network.go
```diff
@@ -0,0 +1,61 @@
...
+	Params:              chaincfg.MainNetParams(),
+	DcrdRPCServerPort:   "9109",
```
davecgh, line 27: Confirmed all settings match their previous values.

internal/config/network.go
jholdstock: Ah interesting! I noticed you making that change to `rpcclient.ConnConfig` in the `newHub` func in the PR decred/dcrpool#356. I wondered what difference it would make, now I know!

I'll update that.

internal/webapi/getfeeaddress.go
jholdstock: I'm not typically too strict about it because I feel in some cases it can harm readability more than it helps. I really don't like it when a line of code becomes a paragraph. So long as its not unreasonable, like 150, I let it slide.

davecgh (approved):

PR #431, opened 2023-09-14
Don't use RPC to determine DCP0010 status

jholdstock:
The activation heights are known for mainnet and testnet, so they can be hard-coded and RPC does not need to be used.

Requires #430

### DIFF
--- a/internal/config/network.go
+++ b/internal/config/network.go
@@ -20,6 +20,9 @@ type Network struct {
 	// DCP0005Height is the activation height of DCP-0005 block header
 	// commitments agenda on this network.
 	DCP0005Height int64
+	// DCP0010Height is the activation height of DCP-0010 change PoW/PoS subsidy
+	// split agenda on this network.
+	DCP0010Height int64
 }
 
 var MainNet = Network{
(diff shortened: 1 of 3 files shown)


internal/config/network.go
```diff
@@ -42,6 +48,9 @@ var TestNet3 = &Network{
 	// DCP0005Height on testnet3 is block
 	// 0000003e54421d585f4a609393a8694509af98f62b8449f245b09fe1389f8f77.
 	DCP0005Height: 323328,
+	// DCP0010Height on testnet3 is block
+	// 00000000c78c04b8b6f7a6a4e3c602abff55f93924e154cac7be3a039941c6cd.
+	DCP0010Height: 867648,
```
davecgh, line 53: This isn't correct.

```sh
$ dcrctl --testnet getblockchaininfo | jq -c '.deployments.changesubsidysplit | {status,since}'
{"status":"active","since":877728}
$ dcrctl --testnet getblockchaininfo | jq .deployments.changesubsidysplit.since | dcrctl --testnet getblockhash -
000000000000c7fd75f2234bbff6bb81de3a9ebbd2fdd383ae3dbc6205ffe4ff
```

davecgh (changes requested):

internal/config/network.go
jholdstock: Damn, looks like https://voting.decred.org/testnet is wrong. dcrdata matches the values you have.

davecgh (approved):

PR #438, opened 2023-09-18
Move signal to external package, improve and reuse.

jholdstock:
- Remove unused code.
- Improve compatibility for operating systems
- Move to internal package. Reuse in v3tool and vote-validator.


dajohi (approved):

cmd/v3tool/main.go
```diff
@@ -125,17 +143,23 @@ func run() int {
...
 		}
 
-		feeAddrResp, err := vClient.FeeAddress(context.TODO(), feeAddrReq, commitmentAddr)
+		feeAddrResp, err := vClient.FeeAddress(ctx, feeAddrReq, commitmentAddr)
 		if err != nil {
+			if errors.Is(err, context.Canceled) {
+				return 0
+			}
 			log.Errorf("getFeeAddress error: %v", err)
 			break
```
davecgh, line 152: I see most of these are `break`ing out to the `return 0` while others, along with the new code, are doing `return 0` directly.  It probably should be consistent.

davecgh (commented):
Nice changes.  It's always good to see properly respecting contexts and OS signals.

cmd/v3tool/main.go
jholdstock: Updated

davecgh (approved):

PR #439, opened 2023-09-18
vspd: Ensure backup loop runs forever.

jholdstock:
Fixing a bug where the backup loop was not a loop - it was only running one time.

### DIFF
--- a/cmd/vspd/main.go
+++ b/cmd/vspd/main.go
@@ -128,14 +128,16 @@ func run() int {
 	// Periodically write a database backup file.
 	shutdownWg.Add(1)
 	go func() {
-		select {
-		case <-ctx.Done():
-			shutdownWg.Done()
-			return
-		case <-time.After(cfg.BackupInterval):
-			err := db.WriteHotBackupFile()
-			if err != nil {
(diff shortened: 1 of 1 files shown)


dajohi (approved):

davecgh (approved):

cmd/vspd/main.go
```diff
@@ -128,14 +128,16 @@ func run() int {
...
 	go func() {
-		select {
-		case <-ctx.Done():
-			shutdownWg.Done()
-			return
-		case <-time.After(cfg.BackupInterval):
-			err := db.WriteHotBackupFile()
-			if err != nil {
-				log.Errorf("Failed to write database backup: %v", err)
+		for {
```
davecgh, line 131: I saw this in the original PR, but thought it was intended to only be a single backup after the interval had expired.  I should've asked about it.

jholdstock: As it was it created a time-bomb. After the backup interval had passed it became impossible to cleanly shutdown the process.

davecgh: Yeah, makes sense.  It would indeed do that because `shutdownWg.Done()` isn't deferred so it wouldn't run.

PR #440, opened 2023-09-18
webapi: Abort requests if web cache not ready.

jholdstock:
A new middleware aborts any requests which require the data cache if the cache is not yet initialized. An explicit error is returned so the admin will be immediately aware what has gone wrong.

Previously, webpages were rendered and JSON responses were sent with zero values, and with no indication of anything being wrong. This was sometimes difficult to notice, and even when noticed the cause was not immediately apparent.

### DIFF
--- a/internal/webapi/admin.go
+++ b/internal/webapi/admin.go
@@ -144,8 +144,10 @@ func (w *WebAPI) statusJSON(c *gin.Context) {
 
 // adminPage is the handler for "GET /admin".
 func (w *WebAPI) adminPage(c *gin.Context) {
+	cacheData := c.MustGet(cacheKey).(cacheData)
+
 	c.HTML(http.StatusOK, "admin.html", gin.H{
-		"WebApiCache":  w.cache.getData(),
+		"WebApiCache":  cacheData,
 		"WebApiCfg":    w.cfg,
 		"WalletStatus": w.walletStatus(c),
 		"DcrdStatus":   w.dcrdStatus(c),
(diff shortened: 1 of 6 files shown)


davecgh (commented):
This is a great improvement.  It's much nicer to get explicit errors.

That said, and it could just be because I'm not super familiar with this code base yet, but why not just store the cache in the `WebAPI` type and require the cache to be initialized its constructor as opposed to stuffing it into the gin context where it has to constantly be checked, needs an additional middleware, etc?

jholdstock:
The cache failing to initialize shouldn't be a total showstopper - the vspd process should continue startup as normal so the rest of its functions can occur (tickets are adding to voting wallets etc). It's a recoverable error - chances are the cache will update sooner or later - e.g. once the RPC connection to dcrd is re-established. 

As for why to use a middleware:
1. The .initialized() check and the associated error handling only have to exist in a single place rather than being duplicated N times in N different handlers (N is 7 currently).
2. Putting it first in the chain of middlewares means it will execute first - before hitting the database or any RPCs, before checking request authentication, etc. Failing faster in this fashion means clients will get their responses faster, and it will also lighten the load on the server.

jholdstock:
Pushed one more commit which attempts to initialize the cache before returning an error.

jholdstock:
Your suggestion is correct, that would work and be safe. And while I do agree that the use of Context has significant drawbacks, in this particular case it actually does offer a small benefit in return. Retrieving from context implies that it must have been added to context - ie. it ensures that the middleware has been run and initialization has been checked. Accessing it directly offers no such guarantee, there would be no hard check that middleware has definitely been executed. 

When context is used, middleware misconfiguration results in a development-time panic. When it isn't, the same misconfiguration results in a "silent failure" which can only be noticed by manual code review.

davecgh:
The panic was actually precisely my concern!  The cache isn't mandatory, but a slight configuration will lead to the entire VSP going down at run time.  However, I guess it wouldn't be too bad since the voting wallets are separate and thus it would only be the front end.

davecgh (approved):

PR #441, opened 2023-09-19
database: Remove RWMutex protecting ticket bucket.

jholdstock:
The mutex was added to the ticket bucket as a pre-emptive safety measure (#150), but the bbolt project documentation seems to indicate that it isn't necessary.

https://github.com/etcd-io/bbolt/blob/v1.3.7/README.md#transactions

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -25,8 +25,6 @@ type VspDatabase struct {
 	db                   *bolt.DB
 	maxVoteChangeRecords int
 	log                  slog.Logger
-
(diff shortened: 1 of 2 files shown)


jholdstock:
Requesting review from @jrick because dcrwallet makes heavy use of bbolt and, as far as I can see, it does not maintain any of its own RW locking.

davecgh (approved):
I used bolt extensively in dcrd prior to leveldb and can confirm the transactions behave as the linked documentation states.

PR #444, opened 2023-09-19
Break up and rename blockConnected function

jholdstock:
Splitting the large vspd.go file into several new files, and the blockConnected function into smaller, more specific functions. 

Beyond the fact that small components are easier to maintain and reason about, this is also a step towards providing more granular (and thus more useful) log/error information,.


davecgh (approved):
I've confirmed that the moved code is the same.  The only changes are function renames, splitting code into separate functions and files, and passing the context and rpc client as parameters.

PR #445, opened 2023-09-19
Refine logging of background processes

jholdstock:
Requires #444

Some tweaks to logging to reduce unnecessary spam, demote some less important messages to debug and promote some more significant ones to info.

### DIFF
--- a/internal/vspd/update.go
+++ b/internal/vspd/update.go
@@ -322,7 +322,7 @@ func (v *Vspd) setOutcomes(ctx context.Context, dcrdClient *rpc.DcrdRPC) {
 			continue
 		}
 
-		v.log.Infof("Ticket %s %s at height %d",
-			dbTicket.Hash, dbTicket.Outcome, spentTicket.heightSpent)
+		v.log.Infof("Ticket %s at height %d (ticketHash=%s)",
+			dbTicket.Outcome, spentTicket.heightSpent, dbTicket.Hash)
 	}
 }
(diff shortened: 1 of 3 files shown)


internal/vspd/update.go
```diff
@@ -254,7 +254,7 @@ func (v *Vspd) addToWallets(ctx context.Context, dcrdClient *rpc.DcrdRPC) {
 					}
 				}
 
-				v.log.Infof("Ticket added to voting wallet (wallet=%s, ticketHash=%s)",
+				v.log.Debugf("Ticket added to voting wallet (wallet=%s, ticketHash=%s)",
```
davecgh, line 257: I would think an admin would like to see this in the logs (without having to enable debug logging) in order to easily be able to determine if it was added properly when a user shows up later making claims they registered a ticket and it was accepted but didn't vote.

jholdstock: Good point. The reason I downgraded this one is because it is logged for each wallet, i.e. 3+ times per ticket. I'm going to leave it at info but replace it with a single line.

jholdstock: Meh, the commit message is wrong, but I plan to squash this down to a single commit before merging it anyway.

davecgh: The change to this looks good.  It's failing due to the move of pluralize though.  Resolving this bit.

jholdstock: Yeah that was dumb. Reverted.

davecgh (approved):

PR #446, opened 2023-09-20
Accumulated minor nits and fixes.

jholdstock:
A bunch of small things I've noticed during testing/reviewing. Nothing big enough to warrant its own PR so I've accumulated them here.


davecgh (commented):
Reviewed through `multi: Use const where possible`.

jholdstock:
Rebased and added one extra commit. Good to go.

davecgh (approved):

PR #452, opened 2023-09-27
docs: Add 1.3.1 release note

jholdstock:


docs/release-notes/release-notes.1.3.1.md
```diff
@@ -0,0 +1,19 @@
+# vspd 1.3.1
+
+This is a patch release of vspd which includes the following changes:
+
+- webapi: Add missed tickets to admin page ([#451](https://github.com/decred/vspd/pull/451)).
+
+Please read the [vspd 1.3.0 release notes](https://github.com/decred/vspd/releases/tag/release-v1.3.0)
```
davecgh, line 7: Is this intentionally against 1.3.0?

davecgh: I would've just expected it to be relative to v1.3.0 for a patch release.

jholdstock: No, I think this is correct as-is:

- Those upgrading from 1.3.0 will only be interested in the note they are currently reading, they can ignore this link.
- Those upgrading from 1.2 will want to review the 1.3.0 note as well as the one they are currently reading, they should follow this link.

davecgh: Alright.  I personally always link the relevant commits for the patch release, so it stood out to me.  Either way works.

davecgh (approved):

PR #453, opened 2023-09-27
Backport and prepare 1.3.1

jholdstock:


davecgh (commented):
Looks good.  I'll wait to approve until the result of the release notes is sorted out though since that might have to be changed if the one being backported is changed.

davecgh (approved):

PR #458, opened 2023-11-10
Add docs for listing on decred.org

jholdstock:
This PR consolidates all VSP documentation into this repo. The process for listing a VSP is not really reading for a general user, so does not really belong on docs.decred.org.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,9 @@ requests, and a number of remote servers which host voting wallets. For more
 information about deploying vspd, check out
 [deployment.md](./docs/deployment.md).
 
+The process for listing a new VSP on [decred.org](https://decred.org/vsp/), and
+consequently in Decrediton, is detailed in [listing.md](./docs/listing.md).
+
 ## Development
 
 ### Test Harness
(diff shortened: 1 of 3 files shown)


jholdstock:
#decred/dcrdocs/1235

xaur:
Minor but common issue was expiring certificates. It may be the norm expected from server operators and I'm not sure it is worth mentioning in `listing.md`.

davecgh (commented):
This reads fine, but I did notice some changes versus the docs its replacing on docs.decred.org.

Most of the changes make sense to, but I'm not sure the bits about sys admin and compilation capabilities should be removed.  That is rather important I think.

I also don't think the requirement for the web front end being separate from the voting wallets should be removed.  I realize it allows one less machine, but it also is like a beacon for attackers to take out one of the wallets because they'll know its IP is one and the same.

jholdstock:
Thanks for the reviews. I figured that "dedicated machine for web front-end" was clear enough, but I guess that isn't strictly the same as having a different IP.

PR #468, opened 2024-05-14
config: Introduce NetworkFromName.

jholdstock:
This moves the mapping logic to a more obvious home and enables it to be reused later.


davecgh (approved):
This also has a side effect of being a bit more robust should the network name ever change or a new network is added since it will now show a proper error if the network selection logic choices diverge from the config option choices.

PR #469, opened 2024-05-17
multi: Explicitly handle help requests.

jholdstock:
Checking for --help as an explicit step before parsing any other configs makes the code more intuitive by removing a convoluted bit of error handling, which happened to be unnecessarily duplicated in three places.

Moving it to a function in the internal package makes it reusable by multiple binaries.

This also enables the IgnoreUnknown option to be used whilst parsing for help, which ensures the presence of --help will always result in the help message being printed. This fixes a minor inconsistency where the help message would be printed if the flag was placed before an invalid config, but placing it after would cause an invalid config error to be written instead. For example, `vspd --help --fakeflag` vs `vspd --fakeflag --help`.

### DIFF
--- a/cmd/vote-validator/main.go
+++ b/cmd/vote-validator/main.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022-2023 The Decred developers
+// Copyright (c) 2022-2024 The Decred developers
 // Use of this source code is governed by an ISC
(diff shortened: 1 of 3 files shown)


davecgh (commented):
This reads well.  I'll circle back after I've tested the equivalent changes made in the dcrd PR.

Issue #465, opened 2024-05-13
Create a new admin binary.


jholdstock:
The vspd binary is intended to be a non-interactive long-lived daemon which serves the VSP website and handles the lifecycle of added tickets. However, it currently includes some one-off admin tasks which are necessarily interactive, e.g. [initializing the database](https://github.com/decred/vspd/blob/cdd02dbd2ec607f3041a36a3d3c83e721c8f586e/cmd/vspd/config.go#L435-L456).

The impact of not having a proper home for this kind of admin task can be observed in #311. Interactive password changing is obviously a nice feature to have, but fitting it nicely into the vspd daemon is awkward.

A dedicated binary for performing admin tasks without the concern of remaining non-interactive would provide an obvious ideal location for this kind of function. It could potentially even mean that tasks such as changing the admin password can be performed without having to stop/start the vspd process, another obvious win.

PR #470, opened 2024-05-18
config: Deprecate --configfile.

jholdstock:
This removes the --configfile option from vspd. It introduced quite a few weird edge cases (and a fair bit of code to deal with them) but afaik nobody actually used it. Note that the --homedir option stays, so it is still possible to run vspd with config in a non-default location if required.

### DIFF
--- a/cmd/vspd/config.go
+++ b/cmd/vspd/config.go
@@ -37,7 +37,6 @@ var (
 	defaultHomeDir        = dcrutil.AppDataDir(appName, false)
 	defaultConfigFilename = fmt.Sprintf("%s.conf", appName)
 	defaultDBFilename     = fmt.Sprintf("%s.db", appName)
-	defaultConfigFile     = filepath.Join(defaultHomeDir, defaultConfigFilename)
 	defaultDcrdHost       = "127.0.0.1"
 	defaultWalletHost     = "127.0.0.1"
 	defaultWebServerDebug = false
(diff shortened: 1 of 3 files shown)


davecgh (commented):
This isn't my project, but I personally disagree with this.  I use config file overrides all the time with shell aliases to do testing.

jholdstock:
This feature was incidentally inherited from dcrd a few years ago via copy & pasting `config.go` and it hasn't been used/tested/maintained since. I understand this feature could be more important in dcrd where specifying a different config file whilst reusing an existing datadir would be desirable, but vspd does not have the same restriction of a multi-gigabyte datadir.

I'd like to drop this behavior so that config creation/management can be more easily refactored into a new purpose built binary (#465). Preserving the behavior while nobody uses it creates extra work unnecessarily.

Every config value can still be overwritten via CLI, so testing with shell aliases is still feasible.

davecgh:
Code looks fine, so I'll approve.  I still disagree with the premise though.  I know it's possible to use a shell alias to specify a bunch of command line options, but it is far more annoying than having separate config files that the aliases point to.  You can modify the config file without having to mess with updating the shell aliases and restarting the shell or rehashing, etc.  I also keep the config files in version control so it's easy to compare them and even go back to older versions of software where the options were different.  Config files are generally just a better choice when it comes to managing config.

davecgh (approved):

PR #471, opened 2024-05-20
 Decouple config loading & log initialization

jholdstock:
Loading config and preparing loggers are two separate concepts which should be handled individually. Nesting one inside the other makes little sense and complicates reusing the code.


database/database.go
```diff
@@ -165,8 +160,6 @@ func CreateNew(dbFile, feeXPub string, log slog.Logger) error {
 		return err
 	}
 
-	log.Info("Database initialized")
```
davecgh, line 168: This seems like a log you wouldn't want to lose.

jholdstock: Per #465, creating the database will soon be in a new dedicated binary, so this is somewhat temporary/irrelevant.

davecgh (approved):

PR #473, opened 2024-05-23
vspd: Move config to internal package.

jholdstock:
This enables the config to be reused in multiple binaries - eg. the upcoming vsp admin binary which will be responsible for writing a default config file for new vspd deployments.

### DIFF
--- a/cmd/vspd/config.go
+++ b/cmd/vspd/config.go
@@ -59,12 +59,32 @@ type vspdConfig struct {
 	HomeDir     string `long:"homedir" no-ini:"true" description:"Path to application home directory. Used for storing VSP database and logs."`
 	ConfigFile  string `long:"configfile" no-ini:"true" description:"DEPRECATED: This behavior is no longer available and this option will be removed in a future version of the software."`
 
-	logPath                                   string
-	dbPath                                    string
+	// The following fields are derived from the above fields by loadConfig().
+	dataDir                                   string
 	network                                   *config.Network
 	dcrdCert                                  []byte
 	walletHosts, walletUsers, walletPasswords []string
 	walletCerts                               [][]byte
+}
+
+func (cfg *vspdConfig) Network() *config.Network {
+	return cfg.network
+}
+
+func (cfg *vspdConfig) LogDir() string {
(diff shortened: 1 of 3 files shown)


cmd/vspd/config.go
```diff
@@ -59,14 +59,34 @@ type vspdConfig struct {
...
+
+func (cfg *vspdConfig) LogDir() string {
+	return filepath.Join(cfg.HomeDir, "logs", cfg.network.Name)
+}
+
+func (cfg *vspdConfig) DatabaseFile() string {
+	return filepath.Join(cfg.dataDir, dbFilename)
+}
+
+func (cfg *vspdConfig) DcrdDetails() (string, string, string, []byte) {
```
davecgh: I don't see any issues, but noting that this and `WalletDetails` are extremely easy to misuse since it's a bunch of the same return types meaning the caller could very easily reverse things.  I've seen this type of thing cause so many bugs over the years...

davecgh (approved):
I don't see anything technically incorrect, so approving.  That said, I did an inline comment you may or may not want to address.

cmd/vspd/config.go
jholdstock: Any suggestions to improve? Would you go with 4 separate accessors each returning a single string?

davecgh: I personally prefer structs for things that return a lot of details.  Then you get named fields that make it much harder to get misuse.  Maybe embed cooked structs in the exported `Config` fields and then then these could just return  e.g. `cfg.walletDetails`.

davecgh: So it would be something like:

```Go
type struct WalletDetails struct {
	Users []string
	Passwords []string
	Hosts []string
	Certs [][]byte
```

Then in `Config`:

`walletDetails WalletDetails`

```Go
func (cfg *Config) WalletDetails() WalletDetails {
	return cfg.walletDetails
}
```

jholdstock: Implemented  that, thanks for the suggestion.

davecgh (approved):
Reviewed updates.  Looks good.

PR #474, opened 2024-05-29
types: Remove Revoked from VSP info response.

jholdstock:
This was marked as deprecated in the previous release and is now being removed.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2022-2023 The Decred developers
+// Copyright (c) 2022-2024 The Decred developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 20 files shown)


davecgh (approved):
I'm not sure what process VSP uses for bumping its module majors, but this is a breaking change to the public API and thus necessitates a new major module version.

The way I handle this over in dcrd is by first bumping the major module version to ensure breaking code never makes it into old module versions even when they're pulled by commit hash.

I'm approving since I don't want to hold it up if that's not how you handle module management.

jholdstock:
Yeah sorry, I was a bit too eager with opening the PR and realized I still needed to make changes. Should be good now.

davecgh:
```diff
$ git diff
diff --git a/run_tests.sh b/run_tests.sh
index 40d98c2..ae48f86 100755
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -14,7 +14,7 @@ go version
 # Run tests on root module and all submodules.
 echo "==> test all modules"
 ROOTPKG="github.com/decred/vspd"
-GORACE="halt_on_error=1" go test -race $ROOTPKG/...
+GORACE="halt_on_error=1" go test -race ./...
```

Issue #461, opened 2024-01-31
hot-swapping the fees xpub on an existing VSP


artikozel:
What would the desirable flow be if a VSP operator , for one reason or another, needed to retire or simply lost control of or access to a wallet that was used to initialise the vspd database?
Would re-initialising the vspd database with a different xpub meant to generate addresses for users to send fees to result in the creation of an essentially new VSP and completely erase the track record and history of the existing VSP?

jholdstock:
The only way to change the xpub currently is to create an entirely new database, which essentially means a totally new VSP. All history, and the keypair used to sign requests/responses, would be lost.

Whilst it's totally possible to add code for xpub changing, the vspd daemon isn't the right place to implement it. Thinking about this problem has inspired #465.

Issue #281, opened 2021-06-24
config: Consider not storing admin pass in plain text.


JoeGruffins:
https://github.com/decred/vspd/blob/e42b1cad330596c243baf4c391dc40c981f4de90/config.go#L66

Per [SEI CERT C Coding Standard](https://wiki.sei.cmu.edu/confluence/display/c/MSC18-C.+Be+careful+while+handling+sensitive+data,+such+as+passwords,+in+program+code) it is best practice not to store plain text passwords in memory or on disk. This could be removed from memory, mostly, by storing the sha256 hash of the admin pass, and constant-time comparing the hashes of incoming passwords. For input on startup, the secure terminal could (optionally) be used which works without a gui to input the pass every start-up.

PR #476, opened 2024-05-30
Introduce vspadmin binary.

### LINKED ISSUE #465: Create a new admin binary.

jholdstock:
vspadmin is a new binary which implements one-off admin tasks which are necessarily interactive and thus do not fit neatly into the long-lived vspd binary.

The initial features added to vspadmin are creating blank databases and default config files, both part of the deployment procedure for a new instance of vspd.

In future this binary can become home for additional features such as changing the admin password (#281, #311) and changing the fee xpub (#461).

Closes #465

### DIFF
--- a/cmd/vspadmin/README.md
+++ b/cmd/vspadmin/README.md
@@ -1,0 +1,30 @@
+# vspadmin
+
+vspadmin is a tool to perform various VSP administration tasks.
+
+## Usage
+
+```no-highlight
+vspadmin [OPTIONS] COMMAND
+```
+
+## Options
+
+```no-highlight
+--homedir=                         Path to application home directory. (default: /home/user/.vspd)
+--network=[mainnet|testnet|simnet] Decred network to use. (default: mainnet)
+-h, --help                         Show help message
+```
+
+## Commands
+
+### `createdatabase`
+
+Creates a new database for a new deployment of vspd. Accepts the xpub key to be
+used for collecting fees as a parameter.
+
+Example:
+
+```no-highlight
+  go run ./cmd/vspadmin createdatabase <xpub>
+```
--- a/cmd/vspadmin/README.md
+++ b/cmd/vspadmin/README.md
@@ -26,5 +26,15 @@ used for collecting fees as a parameter.
 Example:
 
 ```no-highlight
-  go run ./cmd/vspadmin createdatabase <xpub>
+$ go run ./cmd/vspadmin createdatabase <xpub>
+```
+
+### `writeconfig`
+
+Writes a config file with default values to the application home directory.
+
+Example:
+
+```no-highlight
+$ go run ./cmd/vspadmin writeconfig
 ```
--- a/cmd/vspadmin/main.go
+++ b/cmd/vspadmin/main.go
@@ -13,11 +13,13 @@ import (
 	"github.com/decred/dcrd/hdkeychain/v3"
 	"github.com/decred/vspd/database"
 	"github.com/decred/vspd/internal/config"
+	"github.com/decred/vspd/internal/vspd"
 	"github.com/jessevdk/go-flags"
 )
 
 const (
-	dbFilename = "vspd.db"
+	configFilename = "vspd.conf"
+	dbFilename     = "vspd.db"
 )
 
 type conf struct {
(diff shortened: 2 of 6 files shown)


cmd/vspadmin/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+### `createdatabase`
+
+Creates a new database for a new deployment of vspd. Accepts the xpub key to be
+used for collecting fees as a parameter.
+
+Example:
+
+```no-highlight
+  go run ./cmd/vspadmin createdatabase <xpub>
```
davecgh: I personally think `go install` with a path to the repo is much preferable over run like this.  First, you have to clone the repo to be able to do the `go run` like this.  Second, you have to build the bin every time with `go run`.

On the other hand, `go install github.com/decred/vspd/cmd/vspadmin@latest` works without cloning and installs the binary.

The same applies to the other cases too, but I'll just comment on this one.

davecgh (commented):
I reviewed the code and looks fine.

That said, I personally **absolutely detest** running things that require separate applications like an admin binary to configure them before they'll even work.  Software should just work by default and then you can tweak it from there if you want.  It is incredibly annoying to run something for the first time only to have it to tell you that you need to go run something else first when it could just do the setup itself.

I'm not the maintainer of this repo, so it's not my decision and that's fine, but I disagree with this change so much that I won't personally be approving it.

jholdstock:
I really appreciate the review even if you don't agree with the premise, thanks.

> Software should just work by default and then you can tweak it from there...

While I strongly agree with this sentiment, I don't believe it's possible with vspd because there are several things which **must** be manually configured before vspd is able to operate - e.g. an xpub key for collecting fees, connection details for voting wallets.

Do you have any other suggestions for how to solve this problem? I guess it would be possible to add default values which get the process to start without any manual config, but then it would just throw a ton of errors and not actually be able to operate.

I'll also mention that this PR is not making the existing situation worse, perhaps just shining a light on it. The manual steps were already required and they are still required after - the only thing which has changed is which binary needs to be executed.

jholdstock:
@dajohi - That is precisely how it worked before this PR, the functionality to perform the two tasks has just moved from one bin to another.

davecgh:
I'm aware there are several things that have to be configured, such as the fee xpub.  The same type of scenario is true (except for seeds instead of a pubkey) in e.g. dcrwallet, dcrdex, Cryptopower, etc.  Requiring the user to perform some setup, preferably guided by the application itself, is something that is totally expected and often can't be avoided, such as in the case of the xpub.

However, there is a very big difference between them needing to set a config flag, add an entry to a config file, etc, and needing to acquire and run an entirely separate binary.  I know it might not seem like much for we dev types, but having to go compile an entirely separate binary (and with the instructions as presented here, you even need to first go download, install, and configure golang itself to be able to do the `go run` it prescribes) is a pretty steep set of steps when you don't already know how all of that works.

To me, it is a _much_ better user experience to simply complain with a detailed message if things are required.  For example, if you were to try to run vspd and it said "An extended public key is required to be set via the --feexpub option .... etc`.  It's pretty clear what needs to be done and there is no whole other set of hoops to jump through to go find and build another binary, configure that other binary to point at the right config locations, etc, etc.

Since vspd already offers a web-based interface, an even better approach, imo, would be to do it it dcrdex does.  When you run it, it points the user at the web address and provides a graphical setup wizard to do all of the one-time setup things.

jholdstock:
Perhaps the database creation and config writing could stay in the vspd binary, but the original inspiration behind the new binary was implementing xpub changing. I considered adding it to the web UI but I think it is too sensitive of an action to be exposed on the internet, especially when it could be implemented on the server CLI with less effort. Shoe-horning it into the existing CLI flags feels really clumsy and not intuitive.

Its interesting that you mention dcrdex as your example because actually that is where the inspiration for this PR came from.

The dcrdex client does indeed guide the user through an easy GUI, which is obviously correct because it is an application aimed at end users. The server however is composed of [a number of CLI bins](https://github.com/decred/dcrdex/tree/master/server/cmd), notably the main server daemon and a separate admin binary (as well as some other small utilities). Its a more elegant solution than just cramming everything into one binary IMO. 

We have specified in the VSP operator guide ever since dcrstakepool launched that operators are expected to be experienced sysadmins and required to build go from source, so I'm not sure I agree that running a different bin is a heavy burden. To be brutally honest about it, if somebody cant figure out how to run a binary or how to build go from source, we probably don't want them operating a VSP anyway.

davecgh:
Understood.  I mean, you're the repo lead, so it's up to you.  It's just a serious pet peeve of mine having separate bins for basic setup tasks.  Interesting point about DEX, though I'll note that I've never had to run any different bins for DEX despite running it for years at this point.

PR #479, opened 2024-06-06
Combine concepts: xpub and last used index 

jholdstock:
Previously the extended public key and the last used address index for that key were two values stored and handled separately. This PR combines the concepts and handles them as a single unit by wrapping them into a new struct `FeeXPub`.

The underlying format for storing these values in the database does not change. The only change is the interface between the database code and the calling code.


davecgh (approved):
Nice change.  It makes sense to keep them together since they are tightly coupled.

jholdstock:
Rebased on master and implemented suggestion.

davecgh (approved):
Still looks good after the update.

PR #480, opened 2024-06-24
vspadmin: Add retirexpub command

### LINKED ISSUE #461: hot-swapping the fees xpub on an existing VSP

jholdstock:
**Database Upgrade Warning**

This PR adds a new command `retirexpub` to vspadmin. It opens an existing vspd database and replaces the currently used xpub with a new one. A database upgrade was required in order to support this functionality.

Current and historic xpubs are displayed on a new tab in the admin page of the web UI.

Closes #461

### DIFF
--- a/cmd/vspadmin/main.go
+++ b/cmd/vspadmin/main.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/decred/dcrd/dcrutil/v4"
 	"github.com/decred/dcrd/hdkeychain/v3"
+	"github.com/decred/slog"
 	"github.com/decred/vspd/database"
 	"github.com/decred/vspd/internal/config"
 	"github.com/decred/vspd/internal/vspd"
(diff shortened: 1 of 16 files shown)


cmd/vspadmin/main.go
```diff
@@ -45,6 +46,21 @@ func fileExists(name string) bool {
 	return true
 }
 
+// validatePubkey returns an error if the provided key is not for the expected
```
davecgh, line 49: Or if it's invalid too.

database/feexpub.go
```diff
@@ -5,58 +5,95 @@
...
+
+	// Find the active xpub - the one with the highest ID.
+	var highest uint32
+	for id := range xpubs {
+		if id > highest {
+			highest = id
+		}
+	}
+
+	return xpubs[highest], nil
```
davecgh, line 65: I believe the initialization of the database ensures there is at least one xpub, so it probably doesn't matter, but this will crash if there are no xpubs in the database.   It seems like it'd be safer to me to have a check before the loop like:

```Go
if len(xpubs) == 0 {
	return FeeXPub{}, /* no xpub configured error */
}

// Find the active xpub - the one with the highest ID.
var highest uint32
...
```

davecgh (commented):
Nice job overall!  I appreciate the good code documentation to help assert it's doing what was intended and addition of solid test coverage.

database/feexpub.go
jholdstock: I think we already have that scenario covered - `AllXPubs()` will return a "bucket does not exist" error if there are no xpubs in the DB.

cmd/vspadmin/main.go
jholdstock: Updated the comment

database/feexpub.go
davecgh: That's what I meant by the bit about saying I believe the initialization of the database ensures there is at least one.  Generally speaking, just because a bucket exists does not necessarily imply it has anything in it.  It probably doesn't matter too much for vspd, but I'm always super careful about such things in dcrd to avoid crashes.

davecgh (approved):

PR #435, opened 2023-09-15
Consider DCP0012 in VSP fee calculations. 

jholdstock:
Upgrade the dcrwallet dependency to pick up the new version of txrules.StakePoolTicketFee which considers the status of DCP0012 in its fee calculation.

PR #481, opened 2024-06-28
multi: Consider DCP0012 in VSP fee calculations.

jholdstock:
Upgrade the dcrwallet dependency to pick up the new version of txrules.StakePoolTicketFee which considers the status of DCP0012 in its fee calculation.


jholdstock:
This change was made in the past (#435) but needed to be reverted (#454) because clients were not updated to use the new calculation. This change can now go back in as of the dcrwallet/decrediton 2.0 release.

davecgh (approved):

PR #487, opened 2024-08-14
Build/test with go 1.22 & 1.23

jholdstock:


davecgh (approved):

davecgh:
README not updated to match, btw, but the PR itself is fine.

jholdstock:
README seems ok to me, did I miss something?

davecgh:
> README seems ok to me, did I miss something?

You're right.  I saw it still said go 1.21 and 1.22, but I see you are in fact updating it in the second commit.  I bet I only had the first commit of the PR selected when I was reviewing.

PR #489, opened 2024-09-19
build: update linter

dajohi:


client/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/decred/vspd/client/v4
 
-go 1.19
```
davecgh, line 3: Why is this bumped?  Nothing requires it.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/decred/vspd
 
-go 1.20
```
davecgh, line 3: Same.  No reason to bump this.

davecgh (changes requested):

go.mod
dajohi: copyloopvar requires it.

davecgh: It doesn't seem very prudent to me to require a new version of Go for all consumers just to add a linter.

jholdstock:
Merged into #488 and #490

PR #494, opened 2024-10-25
webapi: Footer update time relative not absolute

jholdstock:
Using a relative time rather than absolute requires far less effort for a reader to understand, e.g. compare "25 Oct 2024 11:12:49 UTC" to "3 minutes ago".

As a nice side-effect this also fixes a minor privacy issue where using the full absolute timestamp would expose the server timezone to the public.

### Before
Screenshot from 2024-10-25 12-29-28

### After
Screenshot from 2024-10-25 12-29-43


davecgh (approved):
I'm sad to see a new dependency, but code looks fine.

PR #495, opened 2024-11-01
Minor GUI improvements and bugfixes.

jholdstock:
This PR includes many small improvements to the GUI and some bugfixes. Some examples provided below.

Also included is an upgrade from bootstrap 4.5 to 4.6.2. Nothing major changes in the new version, just some improved compatibility/consistency between browsers.

**Needs squashing before merge.**

Deployed to [testnet-vsp.jholdstock.uk](https://testnet-vsp.jholdstock.uk/) and I'm happy to provide reviewers with the admin password.

---

### Fixed an issue where long pages (such as ticket search result) would allow the viewer to scroll beyond the footer

<details>
  <summary>Example of before</summary>
  
</details>

---

### Fixed footer stretching to full widescreen size rather than staying inside bounds of bootstrap container

<details>
  <summary>Before</summary>
  
</details>

<details>
  <summary>After</summary>
  
</details>

---

### More consistent style for inputs (password prompt, ticket search input)

<details>
  <summary>Before</summary>
  
   
</details>

<details>
  <summary>After</summary>
  
  
</details>

---

PR #496, opened 2024-11-01
webapi: Distinguish between current/old xpub

jholdstock:
Based on #495 

Showing current and old pubkeys under separate headers makes the information easier to parse, and also fixes a bug where the current pubkey was showing a retired timestamp of unix epoch.

<details>
  <summary>Before screenshot</summary>
  
</details>

<details>
  <summary>After screenshot</summary>
  
  
</details>


davecgh (approved):
Always nice to see duplicate code removal.

PR #504, opened 2025-12-18
More strict version checks

jholdstock:
Checks were already in place for RPC version, but this was not sufficient to ensure that the binaries have actually been updated. New checks will now enforce the binary and RPC versions of the directly connected dcrd and dcrwallet, and also the dcrd which is backing the connected dcrwallets.


rpc/dcrwallet.go
```diff
@@ -155,6 +143,31 @@ func (w *WalletConnect) Clients() ([]*WalletRPC, []string) {
 	return walletClients, failedConnections
 }
 
+// version uses version RPC to retrieve the API version of dcrwallet.
+func (c *WalletRPC) version() (*semver, error) {
+	var verMap map[string]dcrdtypes.VersionResult
+	err := c.Call(context.TODO(), "version", &verMap)
```
jrick: instead of introducing even more TODO, why not write this with context support and pass in TODO from the caller, if that is still needed?

PR #512, opened 2026-04-18
Enhance fee transaction sanity checks.

jholdstock:
Confirm the existence of all of the inputs to fee transactions provided by clients. If any of the inputs do not exist, or are not known by the local dcrd instance, an error is returned to the client and the ticket is not registered with the VSP.

Closes #508

### DIFF
--- a/internal/webapi/payfee.go
+++ b/internal/webapi/payfee.go
@@ -142,6 +142,29 @@ func (w *WebAPI) payFee(c *gin.Context) {
 		return
 	}
 
+	// Confirm all inputs of the fee transaction exist.
+	for _, input := range feeTx.TxIn {
+		prevOut := input.PreviousOutPoint
+		txOut, err := dcrdClient.GetTxOut(prevOut.Hash, prevOut.Index, prevOut.Tree)
+		if err != nil {
+			w.log.Errorf("%s: dcrd.GetTxOut for fee tx input failed "+
+				"(ticketHash=%s, output=%s:%d:%d, clientIP=%s): %v",
+				funcName, ticket.Hash, prevOut.Hash, prevOut.Index, prevOut.Tree,
+				c.ClientIP(), err)
+			w.sendError(types.ErrInternalError, c)
+			return
+		}
+		if txOut == nil {
+			w.log.Warnf("%s: Fee tx contains non-existent input "+
+				"(ticketHash=%s, output=%s:%d:%d, clientIP=%s)",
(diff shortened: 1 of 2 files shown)


internal/webapi/payfee.go
```diff
@@ -143,6 +143,29 @@ func (w *WebAPI) payFee(c *gin.Context) {
 		return
 	}
 
+	// Confirm all inputs of the fee transaction exist.
+	for _, input := range feeTx.TxIn {
+		prevOut := input.PreviousOutPoint
+		txOut, err := dcrdClient.GetTxOut(prevOut.Hash, prevOut.Index, prevOut.Tree)
+		if err != nil {
+			w.log.Errorf("%s: dcrd.GetTxOut for fee tx input failed "+
+				"(ticketHash=%s, hash=%s, index=%d, tree=%d, clientIP=%s): %v",
```
davecgh: Not a big deal, so I'm approving regardless, but I'll note that it wasn't immediately clear to me which "hash" it was referring to without looking at the arguments.  So, if I were to see that error message in isolation with even less context, I probably wouldn't immediately know which hash it was referring to.

Dcrd uses verbiage like "output %v referenced from transaction %s:%d either does not exist or has already been spent" and then that output formatting is "hash:index:tree".

Perhaps it would be clearer as:

```suggestion
				"(ticketHash=%s, output=%s:%d:%d, clientIP=%s): %v",
```

Or if you prefer to keep each field named, maybe?

```suggestion
				"(ticketHash=%s, outputHash=%s, index=%d, tree=%d, clientIP=%s): %v",
```

davecgh (approved):

internal/webapi/payfee.go
jholdstock: Nice idea

davecgh (approved):

jholdstock:
I held off on merging this it runs on testnet for a while. I have seen the `Fee tx contains non-existent input` error 3 times in 16 days, investigating why. Pretty sure its just slow network propagation which I guess is unavoidable.

jholdstock:
I still see this error every few days, and in every case the ticket and fee transactions are eventually mined successfully which I think confirms my theory that the root cause is network propagation delay.

Given that every wallet on the network is subject to this delay, every wallet using a VSP could potentially encounter this issue and need to send unnecessary retries and remain online longer than necessary.

Given that the only real benefit of this PR is to give an earlier error to wallets which provide an invalid fee transaction, I don't think the trade-off of inconveniencing every other wallet on the network is worth it.

PR #514, opened 2026-05-21
webapi: Give server time to shutdown.

jholdstock:
Passing an already canceled context into server.Shutdown means it has no time to shutdown.

### DIFF
--- a/internal/webapi/webapi.go
+++ b/internal/webapi/webapi.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2025 The Decred developers
+// Copyright (c) 2020-2026 The Decred developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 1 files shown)


internal/webapi/webapi.go
```diff
@@ -153,7 +153,11 @@ func (w *WebAPI) Run(ctx context.Context) {
 		<-ctx.Done()
 
 		w.log.Debug("Stopping webserver...")
-		_ = w.server.Shutdown(ctx)
+
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
davecgh, line 157: I'm not super familiar with the shutdown code, but this seems like it could potentially stop the shutdown early before it's actually done.  That could be an issue if there is a database involved and it needs to flush data to disk.

davecgh: Ok, I just looked at the code and it's actually the http server here.  So, no issues.

davecgh (approved):

internal/webapi/webapi.go
jholdstock: Yep this is just for shutting down the webserver, and there should never be request which takes anywhere near 10 seconds to process. I think without this change the server was being shutdown instantly with in-progress requests just being interrupted, so this is a huge improvement

PR #517, opened 2026-06-15
Mitigate DoS and fix dcrd orphan error handling

jholdstock:
First two commits look huge but they are mostly indentation changes.

Closes #509

### DIFF
--- a/rpc/dcrd.go
+++ b/rpc/dcrd.go
@@ -10,7 +10,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"strings"
+	"regexp"
 
 	"github.com/decred/dcrd/blockchain/standalone/v2"
 	"github.com/decred/dcrd/chaincfg/chainhash"
@@ -29,10 +29,11 @@ const (
 	// we dont need to import the whole package.
 	ErrRPCDuplicateTx = -40
 	ErrNoTxInfo       = -5
-	// This error string is defined in dcrd/internal/mempool. Copied here
-	// because it is not exported.
-	ErrUnknownOutputs = "references outputs of unknown or fully-spent transaction"
 )
+
+// ErrOrphan error string is defined in dcrd/internal/mempool. Copied here
+// because it is not exported.
+var ErrOrphan = regexp.MustCompile(`orphan transaction \w+ references output \w+:\d+ of unknown or fully-spent transaction`)
 
 // DcrdRPC provides methods for calling dcrd JSON-RPCs without exposing the details
 // of JSON encoding.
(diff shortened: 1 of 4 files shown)


rpc/dcrd.go
```diff
@@ -29,11 +29,12 @@ const (
...
 	ErrRPCDuplicateTx = -40
 	ErrNoTxInfo       = -5
-	// This error string is defined in dcrd/internal/mempool. Copied here
-	// because it is not exported.
-	ErrUnknownOutputs = "references outputs of unknown or fully-spent transaction"
 )
 
+// ErrOrphan error string is defined in dcrd/internal/mempool. Copied here
+// because it is not exported.
+var ErrOrphan = regexp.MustCompile(`orphan transaction \w+ references output \w+:\d+ of unknown or fully-spent transaction`)
```
davecgh, line 36:
```suggestion
var ErrOrphan = regexp.MustCompile(`orphan transaction [[:xdigit:]]+ references output [[:xdigit:]]+:\d+ of unknown or fully-spent transaction`)
```

It doesn't really matter, but that is technically more accurate because they'll always be hashes in hex.

davecgh (commented):
👍 on reversing the error checks.  I know it's mostly just a style thing, but I very much prefer the fail fast style over burying the core logic and errors in layers of indentation.

Not only does it read better (to me anyway), but I also find it easier to reason about.  "Bad stuff is out of the way, now to the positive path" versus "some good stuff here, but, oh, if something bad in that path, well, oh, but first the good stuff in that path, and then, ...".  It is similar to why a lot of people like exceptions.

e.g:

```Go
if bad {
	/* return error */
}
if bad2 {
	/* return error2 */
}
/* do the core logic */
```

over

```Go
if good {
	/* do some core logic */
	if good2 {
		/* do some more core logic */
	} else {
		/* return error2 */
	}
} else {
	/* return error */
}
```

While computers are fast enough that it doesn't really make a big difference, especially when it's not in the hot path, not only is the former style much easier to read (imo), it is also almost always slightly faster due to the way the branch predictor works in modern processors.

Also, thanks for splitting out the wrapper and logic change into separate commits to make it easier to review.

rpc/dcrd.go
jholdstock: Nice, I didn't know regex in go supported that kind of bracketed token/placeholder (I'm also not sure of the proper term to describe `\w`, `\d` etc 😆). 

I'm going to leave it as-is for now because I've already validated this by running it on https://testnet-vsp.jholdstock.uk for a week, but will remember it for next time.

davecgh: Makes sense.  By the way, in case you ever want to search them, the `\w`, `\d` etc are called meta sequences.  The `[a-fA-F0-9]`, `[[:xdigit:]]`, etc are called character classes.

More generally, all recognized "items" of the regular expression are called tokens.

PR #518, opened 2026-06-22
ci: Update linter to 2.12.2.

jholdstock:
Also:
- Remove goconst linter because its latest release is far too sensitive by default, and it doesn't offer enough value to be worth manual configuration.
- Pull the linter install script from the branch "main" rather than "master" as notified in the 2.12.0 release note https://github.com/golangci/golangci-lint/releases/tag/v2.12.0


davecgh (approved):
Oh lovely.  Cause a ton of churn over a useless name change.

PR #522, opened 2026-08-19
rpc: Log connected dcrd version

jholdstock:


rpc/dcrd.go
```diff
@@ -121,22 +121,22 @@ func (d *DcrdConnect) Client() (*DcrdRPC, string, error) {
...
+func (c *DcrdRPC) checkVersion() (string, error) {
 	var verMap map[string]dcrdtypes.VersionResult
 	err := c.Call(context.TODO(), "version", &verMap)
 	if err != nil {
-		return err
+		return "", err
 	}
 
-	return errors.Join(
+	return verMap["dcrd"].VersionString, errors.Join(
```
davecgh: This can panic if it ever gets back a response without that key.

jholdstock: Fixed

davecgh (approved):

PR #525, opened 2026-08-21
Add a couple of missing log values.

jholdstock:


davecgh (approved):
Those would definitely show up in the logs all messed up!

PR #530, opened 2026-08-21
docs: Add 1.5.0 release note

jholdstock:


docs/release-notes/release-notes.1.5.0.md
```diff
@@ -0,0 +1,111 @@
...
+   voting at all times. On each server:
+    1. Stop dcrwallet.
+    1. Stop dcrd.
+    1. Install new dcrd binary and start.
+    1. Wait for any dcrd database upgrades to complete.
+    1. Check dcrd log file for warnings or errors.
+    1. Install new dcrwallet binary and start.
+    1. Wait for any dcrwallet database upgrades to complete.
+    1. Check dcrwallet log file for warnings or errors.
+1. Ensure dcrd on the vspd server has completed all database upgrades.
```
davecgh, line 44: Not a big deal, but this and the next one are repeats.  e.g. It already says above to wait for any dcrd database upgrades to complete.

jholdstock: I think one of us must be mis-reading because I don't see the repetition. Note that the actions in the nested list are happening on the voting wallet host and the main list is the vspd host. Do you think that needs to be made more clear?

davecgh: I suppose it's fine since they're numbered lists that are taking place on different servers.

I just see:

```
... On each server:
...
  - Wait for any dcrd database upgrades to complete.
  - Check dcrd log file for warnings or errors.
  ...
- Ensure dcrd on the vspd server has completed all database upgrades.
- Check dcrd log file for warnings or errors.
```

Which stands out as repetitive to me, but I see your point.

davecgh (approved):

PR #534, opened 2026-08-30
Embed web assets by default.

jholdstock:
Public web assets and HTML templates are now embedded in a compiled vspd binary and used by default, enabling the binary to be portable - it can be run from any location rather than only from the root of this repository. This is particularly useful when wishing to build the binary on one machine and run it on another.

For admins who depended on the old behaviour, e.g. those running with customized templates, loading from the source directory can be re-enabled using a new config flag `--dontembedassets`.

Assets are always loaded from the source dir when `--webserverdebug` is enabled.


davecgh (approved):
Looks good and a very nice improvement.  Portable binaries are a huge strength of Go.

